PR #1, opened 2021-11-04
Add arena background color variable

l-Luna:
Adds a bgColor field to arenas that allows specifying their background colour.


mods/example/scripts/battles/waves/vironeedle.lua
```diff
@@ -4,6 +4,9 @@ function Vironeedle:onStart()
     local attackers = self:getAttackers()
 
     local ratio = self:getEnemyRatio()
+    
+    Game.battle.arena.color = {1, 1, 1}
+    Game.battle.arena.bgColor = {0.2, 0.2, 0}
```
NyakoFox, line 9: You should probably revert your test changes here ^^;;

src/object/game/arenasprite.lua
```diff
@@ -23,7 +23,7 @@ end
 
 function ArenaSprite:draw()
     if self.background then
-        love.graphics.setColor(0, 0, 0)
+        love.graphics.setColor(self.arena.bgColor[1], self.arena.bgColor[2], self.arena.bgColor[3])
```
NyakoFox, line 26: This function takes a table, so you can just do `love.graphics.setColor(self.arena.bgColor)`. That'd also allow it to support alpha, which this doesn't do

src/object/game/arena.lua
```diff
@@ -6,6 +6,7 @@ function Arena:init(x, y, shape)
     self:setOrigin(0.5, 0.5)
 
     self.color = {0, 0.75, 0}
+    self.bgColor = {0, 0, 0}
```
NyakoFox, line 9: The naming scheme for variables is `snake_case`, so `self.bg_color` is better here. Also, all objects have `setColor` and `getColor`, so I'd recommend implementing `setBackgroundColor` and `getBackgroundColor` for the arena.

NyakoFox (changes requested):
Looks good, I just have a few nitpicks

l-Luna:
should all be fixed

NyakoFox:
LGTM, thanks!

PR #2, opened 2021-11-13
Add more image layer options for backgrounds

l-Luna:
Adds some custom properties for use with image layers: `speedX` and `speedY` (numbers) set the speed of the image layer, `wrapX` and `wrapY` (bools) determine whether the texture should wrap horizontally and/or vertically, and `fitToScreen` (bool) extends the image to fit the screen (stretching it or repeating it based on `wrapX/Y`). `scaleX` and `scaleY` (numbers) set the scale of the image.

These allow for having scrolling images in the background, without potentially needing to manually repeat or scale an image to fit.

### DIFF
--- a/src/object/game/world.lua
+++ b/src/object/game/world.lua
@@ -321,6 +321,7 @@ function World:loadImage(layer, depth)
         sprite.width = SCREEN_WIDTH
         sprite.height = SCREEN_HEIGHT
     end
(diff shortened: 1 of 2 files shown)


NyakoFox:
Just a tiny nitpick, can you change the property names to flat case to stay consistent with the rest of the systems? (ex. `speedx` instead of `speedX`)

l-Luna:
fixed

PR #5, opened 2022-02-06
Added setGameTile/getGameTitle and setGameIcon/getGameIcon

FriskTheFallenHuman:
This PR allows mods to change the Game's Title bar and Icon from the `mod.lua` file

### DIFF
--- a/mods/_testmod/mod.lua
+++ b/mods/_testmod/mod.lua
@@ -19,6 +19,13 @@ function Mod:init()
 end
 
 function Mod:preInit()
+    -- Set the game title to something different
+    Game:setGameTitle("Example Mod")
+
+    -- Set the windows icon to something different
+    -- For mods please use Assets.getTextureData has argument
+    Game:setGameIcon( Assets.getTextureData( "kristal/banana_1" ) )
+
     -- make kris woobly
     --[[Utils.hook(ActorSprite, "init", function(orig, self, ...)
         orig(self, ...)
--- a/src/engine/game/game.lua
+++ b/src/engine/game/game.lua
@@ -542,6 +542,26 @@ function Game:getSoulPartyMember()
     return current
 end
 
+function Game:setGameTitle(title)
+    love.window.setTitle(title)
+end
+
+function Game:getGameTitle()
+    return love.window.getTitle()
+end
+
+function Game:setGameIcon(iconData)
+    -- Make sure our icon isn't null.
+    if iconData == nil then return end
+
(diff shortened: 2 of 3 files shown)


src/engine/game/game.lua
```diff
@@ -542,6 +542,26 @@ function Game:getSoulPartyMember()
...
+    love.window.setTitle(title)
+end
+
+function Game:getGameTitle()
+    return love.window.getTitle()
+end
+
+function Game:setGameIcon(iconData)
+    -- Make sure our icon isn't null.
+    if iconData == nil then return end
```
NyakoFox, line 555: i dont know if this is required, if someone passes bad data then it _should_ error

src/engine/vars.lua
```diff
@@ -1,11 +1,13 @@
  -- replaces mod list with "Start game", should be a string of the mod id
 TARGET_MOD = nil
-
+TARGET_ICON = "icon.png"
```
NyakoFox, line 3: why did you add unused variables here?

src/engine/game/game.lua
```diff
@@ -542,6 +542,26 @@ function Game:getSoulPartyMember()
     return current
 end
 
+function Game:setGameTitle(title)
```
NyakoFox, line 545: maybe its a good idea to change "Game" to "Window" in these functions? ex. `setWindowTitle()`

mods/_testmod/mod.lua
```diff
@@ -19,6 +19,13 @@ function Mod:init()
 end
 
 function Mod:preInit()
+    -- Set the game title to something different
+    Game:setGameTitle("Example Mod")
+
+    -- Set the windows icon to something different
+    -- For mods please use Assets.getTextureData has argument
+    Game:setGameIcon( Assets.getTextureData( "kristal/banana_1" ) )
```
NyakoFox, line 27: This isn't where `Assets.getTextureData` should be, it should be handled in the function itself

NyakoFox (changes requested):

DimitriBarronmore:
I feel like I have to point out the obvious here and say that this isn't necessary. The only one of these functions which isn't a single line arguably should have been, and nothing stops the modder from writing those single lines themselves.

Instead of adding pointless aliases which could essentially be replaced with a variable pointing to the original function, a more useful step towards making it simple for modders to change the icon and title would just be to make sure they reset when the game transitions back to the menu. This doesn't solve that or any problems, it just makes things more confusing.

PR #8, opened 2022-04-09
Fix inputs not resetting on crash

Archie-osu:
Hopefully covers all edge-cases, but it's never too late to fix it anyway 😄


NyakoFox:
lgtm

PR #10, opened 2022-07-01
Tweak example mod to not spawn the wall guardians if Susie never hit the wall

prokube:
I make all the important fixes, I know


NyakoFox:
lgtm

PR #22, opened 2023-02-28
Adding mirrors to Kristal

AcousticJammYT:
Mirrors only have one property: "offset". The higher the number, the further down the reflection will draw.


vitellaryjr:
implemented this code as a separate commit instead of using pull request, issue resolved: https://github.com/KristalTeam/Kristal/commit/bd63f4cdca541b94e290d175fa93384e726f0e68

PR #32, opened 2023-07-28
Add Event interaction buffer

MrOinky:
In DELTARUNE, there is a small buffer period after finishing an interaction where the player cannot interact with that object (and theoretically any other objects), this adds that buffer into Kristal.

### DIFF
--- a/src/engine/game/world/event.lua
+++ b/src/engine/game/world/event.lua
@@ -29,6 +29,9 @@ function Event:init(x, y, w, h)
 
     -- Sprite object, gets set by setSprite()
     self.sprite = nil
+
+    -- Duration that the player cannot interact with events for on finishing interaction (in frames, at 30fps)
+    self.buffer = 5
 end
 
 --[[ OPTIONAL FUNCTIONS
--- a/src/engine/game/world/events/npc.lua
+++ b/src/engine/game/world/events/npc.lua
@@ -36,6 +36,8 @@ function NPC:init(actor, x, y, properties)
     self.set_value = properties["setvalue"]
 
     self.interact_count = 0
+
+    self.buffer = 5
 end
 
 function NPC:onInteract(player, dir)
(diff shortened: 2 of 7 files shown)


src/engine/game/world/event.lua
```diff
@@ -29,6 +29,9 @@ function Event:init(x, y, w, h)
 
     -- Sprite object, gets set by setSprite()
     self.sprite = nil
+
+    -- Duration that the player cannot interact with events for on finishing interaction (in seconds)
+    self.buffer = (5/30)
```
SylviBlossom, line 34:
```suggestion
    self.interact_buffer = (5/30)
```

src/engine/game/world/events/npc.lua
```diff
@@ -36,6 +36,8 @@ function NPC:init(actor, x, y, properties)
     self.set_value = properties["setvalue"]
 
     self.interact_count = 0
+
+    self.buffer = 5
```
SylviBlossom, line 40:
```suggestion
    self.interact_buffer = (5/30)
```
Was not adjusted to be in seconds

src/engine/game/world/player.lua
```diff
@@ -116,6 +122,7 @@ function Player:interact()
     table.sort(interactables, function(a,b) return a.dist < b.dist end)
     for _,v in ipairs(interactables) do
         if v.obj:onInteract(self, self.facing) then
+            self.interact_buffer = v.obj.buffer or 0
```
SylviBlossom, line 125:
```suggestion
            self.interact_buffer = v.obj.interact_buffer or 0
```

SylviBlossom (commented):
Would probably be best to name the event variable something like `interact_buffer` instead of just `buffer` to be more specific, since it's a class others extend. Otherwise looks good

MrOinky:
Yeah, `interact_buffer` does seem more suitable given the nature of the class, and is consistent with the variable name used for `player`, went ahead and applied the suggested changes

PR #34, opened 2023-09-06
icons and soulshine animation

Idunno1:
actions that involve multiple party members now change everyone's icons. sped up the soul shine animation in power acts


launch.bat
Idunno1, line 1: forgot to ignore this

PR #53, opened 2023-11-17
Fixed Weird Noelle offset and sprites where needed

FireRainV:
Still missing the snowglobe winning animation


NyakoFox:
Merged manually, thanks!

PR #55, opened 2023-11-19
Remove duplicates from the options menu

Agent517:
Doesn't matter all that much, but its an oddity I noticed.

Fullscreen, Window Scale, Border, and Simplify VFX were in both the general and graphics menu, so these were removed from the general menu.


SylviBlossom:
This was intentional, "general" was meant to be the quick access category, with the most immediately useful options from other categories

PR #61, opened 2023-11-30
Merged all Kristal test mods

FireRainV:
Reduce bloat in the mod menu.
Few fixes to the test mod.


mods/example/mod.json
```diff
@@ -13,5 +13,7 @@
     "inventory": {
         "items": ["darkburger", "darkburger", "ultimate_candy"],
         "key_items": ["cell_phone"]
-    }
+    },
+    
+    "hidden": true
```
NyakoFox, line 18: This shouldn't be hidden.

FireRainV: Just undid that

PR #65, opened 2023-12-03
Toggle Noclip debug hotkey

FireRainV:
Ctrl+N to toggle noclip in the world or battles


NyakoFox:
Please make different PRs for different things. Anyway, the reason why we don't use the tension function is because it caps the tension to 100%, while the debug key was for 200% (to test Snowgrave).

PR #73, opened 2023-12-08
Fixed inaccurate Chapter 1 Power Menu Vertical Line

DiamondBor:
In chapter 2, the vertical line goes off-screen because "Why not?" said Toby. in chapter 1, it doesnt

but in kristal, in both chapter 1 and 2, it goes off-screen

so i changed that

### DIFF
--- a/src/engine/game/world/ui/dark/darkpowermenu.lua
+++ b/src/engine/game/world/ui/dark/darkpowermenu.lua
@@ -192,7 +192,11 @@ function DarkPowerMenu:draw()
 
     Draw.setColor(PALETTE["world_border"])
     love.graphics.rectangle("fill", -24, 104, 525, 6)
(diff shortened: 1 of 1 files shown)


src/engine/game/world/ui/dark/darkpowermenu.lua
```diff
@@ -192,7 +192,11 @@ function DarkPowerMenu:draw()
 
     Draw.setColor(PALETTE["world_border"])
     love.graphics.rectangle("fill", -24, 104, 525, 6)
-    love.graphics.rectangle("fill", 212, 104, 6, 200)
+    if Game:getConfig("oldUIPositions") == true then
```
NyakoFox, line 195: `== true` is not needed.

DiamondBor: ok i changed it

PR #76, opened 2023-12-08
Light World Money starts from 0 instead of 2

FireRainV:
This isn't Deltarune, no need for starting money. Let the players customize it as they want.


NyakoFox:
People can just set it themselves if they need to in one of the mod load callbacks.

FireRainV:
I thought it may be better to initialize it as 0 as it's probably the better starting point.

PR #77, opened 2023-12-08
Added Game.all_party

FireRainV:
This returns all defined party members, even ones who are not in your current party. Save points will heal all defined party member, just like in Deltarune.


FireRainV:
Already exist as `Game.party_data`

PR #82, opened 2023-12-10
Give Item debug menu now shows ID and proper description for light items

FireRainV:
The ID instead of name is for ease of knowing what is what for items with the same name. (Like the Egg)
Also, it's consistent between the rest of the menus.


FireRainV:
I can revert it if you want, but currently, it just displays "Example Description" for all light world items.
It isn't perfect, since you can see stuff like `[wait:5]` and such. Tell me if you want me to revert it.

FireRainV:
Okay, so instead of showing a description at all, it will display "Give this item." for dark items and "Give this light item." for light items. I think it's better that way. It's also similar to other debug menus.

NyakoFox:
I don't really agree to these changes, plus the commit history is a massive mess... I'll have to close this, sorry.

PR #88, opened 2023-12-15
Accuracy fix for DarkMenu, HealthBar, BattleUI and ActionBox transition animations

Scarm2:
This PR implements code-accurate animations for dark UI.

DarkMenu, HealthBar & BattleUI transitions now use the rounded lerp functions found in Deltarune, although I have made them into floor/ceil functions as they break on framerates > 30fps if rounded.

ActionBox easing functions have been replaced by the weird fake lerp found in Deltarune, with a minor fix which was an oversight in Deltarune itself (check comment)

ActionBox's `data_offset` now also almost matches the Deltarune one ("almost" because the out transition cannot be done because lack of "spaghetti code" + see comment on line 162 in `battleui.lua`, however when transitioning in, it is code-accurate)

### DIFF
--- a/src/engine/game/battle/ui/actionbox.lua
+++ b/src/engine/game/battle/ui/actionbox.lua
@@ -4,8 +4,7 @@ local ActionBox, super = Class(Object)
 
 function ActionBox:init(x, y, index, battler)
     super.init(self, x, y)
-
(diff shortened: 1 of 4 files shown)


NyakoFox:
Thanks!

PR #95, opened 2023-12-31
mod.json {"useSaves": "has_saves"} skips file select if no saves

skarph:
ive seen some users wanting this (like, two i think), and the only way to do it was through `preview.lua` abuse


skarph:
just realized `mod["useSaves"] == true` might break things (in theory), nevermind...

PR #99, opened 2023-12-31
Re-enters the shop you were in upon quick reload

FireRainV:


src/kristal.lua
```diff
@@ -957,7 +958,10 @@ function Kristal.quickReload(mode)
                         -- If we had an encounter, restart the encounter
                         if encounter then
                             Game:encounter(encounter, false)
+                        elseif shop then -- If we were in a shop, re-enter it
+                            Game:enterShop(shop)
                         end
+
```
Dobby233Liu, line 964: why the newline

PR #101, opened 2024-01-02
onMapBorder mod call

DiamondBor:
Like Mod:onMapMusic, but for borders. An example of why someone would want this is if they have a map that has two different 'areas' with different borders, that change as they advance through the map.

Code example:
```
function Mod:onMapBorder(map, border)
    if map.name == "name" and (border == "cyber") then
        if Game:getFlag("flag", false) == true then
            return "city"
        end
        return border
    end
end
```


NyakoFox:
This PR accidentally includes changes from another PR of yours.

PR #105, opened 2024-01-05
Display ID instead of name

FireRainV:
Also fix description and make is simpler.


NyakoFox:
Resolved with a3514e7103f67ae6bc6ee622341655eb1bc5d286.

PR #110, opened 2024-01-20
Make sure tempsave is immediately loaded on quick reload

Dobby233Liu:
This prevents the postInit event from being called with is_new_file = true upon a temp load.

Adds the `Game:enter(last, save:table, save_id:number)` overload for this purpose


Dobby233Liu:
Ironically the fix doesn't apply to DP... probably means this (if approved) should be backported there

PR #119, opened 2024-02-23
Added getPartyCharacter function for only the party train

FireRainV:
Also fixes issues with the world soul


NyakoFox:
Manually merged.

PR #123, opened 2024-02-26
transition event sound properties

HUECYCLES:
adds:

- reverse_splat.wav
- self.lancer and self.lancer_exit bool properties
- self.door bool property
- self.sound and self.pitch string properties


NyakoFox:
I implemented it in my fangame using two properties, one for when you touch the transition and one for when you're about to go to the next map.

PR #131, opened 2024-04-03
Light World LV save to file

FireRainV:
I need it for something in magical-glass, it may be used later on for light world saves.


FireRainV:
Maybe in another time, I can't really find a good use for this.

PR #135, opened 2024-05-15
Mod-defined custom soul colors in mod select

AcousticJammYT:
To use it, just define `"soulColor": [1, 0, 0]` in your `mod.json` file.

### DIFF
--- a/src/engine/objects/gonerchoice.lua
+++ b/src/engine/objects/gonerchoice.lua
@@ -29,7 +29,10 @@ function GonerChoice:init(x, y, choices, on_complete, on_select)
 
     self.soul = Sprite("player/heart_blur")
     self.soul:setScale(2, 2)
-    self.soul:setColor(1, 0, 0)
+    self.soul:setColor(Kristal.getSoulColor())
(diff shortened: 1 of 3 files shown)


src/engine/objects/gonerchoice.lua
```diff
@@ -29,7 +29,10 @@ function GonerChoice:init(x, y, choices, on_complete, on_select)
 
     self.soul = Sprite("player/heart_blur")
     self.soul:setScale(2, 2)
-    self.soul:setColor(1, 0, 0)
+    self.soul:setColor(Kristal.getSoulColor())
+	if MainMenu.mod_list:getSelectedMod().soulColor then	-- Add Game detection
```
NyakoFox, line 33: This'll trigger if you're in game, not just the menu, won't it?

NyakoFox: No, because you're immediately overwriting it with the `setColor` in the if statement.

AcousticJammYT: oh shit you're right, that's a thing I forgot to add, hang on a sec

PR #140, opened 2024-05-26
Semi-Complete Recruit System

FireRainV:
This time, it gets the enemy instead of creating it, which prevents any possible crash.


FireRainV:
I've almost finished it, just need to figure out how to create a new file for the recruits.

PR #142, opened 2024-05-27
Recruit System - FULLY COMPLETED

FireRainV:
Uses it's own file separate from enemy battler.


NyakoFox:
LGTM, thanks!

FireRainV:
It was fun doing it, makes me better at understanding how Kristal works.

PR #144, opened 2024-05-28
Fixed issues related to the recruit system

FireRainV:


mods/_testmod/scripts/data/recruits/starwalker.lua
```diff
@@ -6,9 +7,14 @@ function Starwalker:init()
     -- Display Name
     self.name = "Starwalker"
     
-    -- How many times an enemy needs to be spared to be recruited.
+    -- self.name = "Werewerewire"
```
NyakoFox, line 10: Why's this here?

src/engine/game/battle/enemybattler.lua
```diff
@@ -654,9 +654,7 @@ function EnemyBattler:defeat(reason, violent)
             end
             self:setRecruitStatus(false)
         end
-        -- if self.done_state == "KILLED" or self.done_state == "FROZEN" then
-            -- Game.battle.xp = Game.battle.xp + self.experience
-        -- end
+        Game.battle.xp = Game.battle.xp + self.experience
```
NyakoFox, line 657: What's the point of this change in regards to the recruit system?

src/engine/game/battle/enemybattler.lua
```diff
@@ -673,7 +671,6 @@ function EnemyBattler:defeat(reason, violent)
     end
     
     Game.battle.money = Game.battle.money + self.money
-    Game.battle.xp = Game.battle.xp + self.experience
```
NyakoFox, line 676: Same here.

FireRainV: it isn't, but it seemed weird that you can gain EXP from sparing enemies, even though it's unused in Deltarune currently.
I can revert it if you want.

mods/_testmod/scripts/data/recruits/starwalker.lua
FireRainV: mostly for testing stuff related to long names, to be accurate with Deltaune. It's commented.
The virovirokun should be used as a base.

src/engine/game/battle/enemybattler.lua
NyakoFox: It's only unused in Deltarune because XP is always set to 0. This works exactly like Deltarune does. Deltarune rules aren't necessarily the same as Undertale, so we shouldn't make assumptions in the engine as such.

FireRainV: I see, I didn't know that. I will revert that change then.

FireRainV: Done

PR #148, opened 2024-05-29
Added all the missing key items

FireRainV:
Includes: BrokenCake, Lancer, Rouxls Kaard and Starwalker.


NyakoFox:
They were left out for a reason, or at least most were -- I don't remember why we left out BrokenCake. Anyway, how was Lancer ported? Doesn't he change according to plot values, and route? Is this customizable?

FireRainV:
For description text, you can just override Item:getDescription().
As for animation, I just made it that you can choose the animation by setting the LancerGotchi.movecon to a value above 10.
11 is stone, 12 is instant sleeping.

FireRainV:
Here's an example:

image

FireRainV:
Just made it even easier to choose an animation:
item:getCustomLancerAnimation() return <animation>

PR #153, opened 2024-05-30
Invulnerability will carry over between waves, like in Chapter 1

FireRainV:
It's a config option that default to true in ch1. It even carry over between battles for some reason :/


NyakoFox:
hell yeah

PR #161, opened 2024-05-30
More Chapter 1 UI Accuracy

FireRainV:


NyakoFox:
The pixel was technically a part of the "TP" sprite, IIRC.

FireRainV:
I can't seem to find that sprite, but even if it is, it basically draws the same thing.

FireRainV:
Done, it now uses the sprite instead of drawing it manually

PR #164, opened 2024-05-30
cutscene + transition = softlock | fixed!

FireRainV:


NyakoFox:
IIRC this is supposed to be so you can have cross-map cutscenes.

FireRainV:
The cutscene won't even play though when transitioning, it will simply softlock you.
It would be smarter to start a new cutscene after going to the next map.

AcousticJammYT:
Not true; I have cross-map cutscenes and they work well in Kristal.

FireRainV:
So what do I do? Should I close this PR?

AcousticJammYT:
Yeah, that probably seems like the best option

FireRainV:
ok

PR #165, opened 2024-05-31
The Dark Inventory is no longer attached to the ball of junk + Stack Overflow when saving fix

FireRainV:
Fixes a bug where you would lose your key items and storage when thrown away.


SylviBlossom:
what is the point of the `dark_item` and `light_item` saving changes? they dont seem tied to the ball of junk and serve an important purpose of maintaining item state and position between dark and light inventories, so merging this now would cause issues for items which do convert between inventories

FireRainV:
> what is the point of the `dark_item` and `light_item` saving changes? they dont seem tied to the ball of junk and serve an important purpose of maintaining item state and position between dark and light inventories, so merging this now would cause issues for items which do convert between inventories

It causes a stack overflow, many people had issues with it.

FireRainV:
just won't work...

PR #167, opened 2024-06-01
Attacking is now way more accurate to Deltarune

FireRainV:
In Kristal, after all enemies were defeated and you still had more attack bars, their afterimage would go on indefinitely, now fixed.

Missing an attack now occurs at the same time as in Deltarune.

Just like in Deltarune, you will only be able to crit when the bar is fully inside the bolt target.


NyakoFox:
> Just like in Deltarune, you will only be able to crit when the bar is fully inside the bolt target.

The problem is, when you have higher FPSes, it makes the window harder and harder to hit. Ours should still be accurate at 30FPS.

FireRainV:
it creates more issues than it solves

PR #170, opened 2024-06-09
Allow mods to specify a custom window icon

Dobby233Liu:
We already has builtin support to set the window title to a mod's name, so there is no reason this shouldn't exist


Dobby233Liu:
Thought this didn't work, i guess DP isn't a great test case

PR #171, opened 2024-06-09
Handle `.git` being a file instead of a folder

Dobby233Liu:


NyakoFox:
didnt know this could happen thats kinda dumb

PR #172, opened 2024-06-09
Itemized events

skarph:
/src/engine/vars.lua now has a fully documented table `KRISTAL_EVENT` of every string used by `Kristal.callEvent`. Every event string passed to `Kristal.callEvent` references this table


NyakoFox:
she is not beating the merge branch main allegations

image

PR #175, opened 2024-06-16
New targeting options for items and spells | XACT code improvement

FireRainV:
You can now specify "any" or "all" as a target, allowing you to target both the party and the enemies with the same item/spell.

XACT is now a state reason of ENEMYSELECT


FireRainV:
@NyakoFox is this any/all feature okay? If not, should I still merge the XACT being a state reason of ENEMYSELECT?

FireRainV:
tbh, maybe this feature would be better when Deltarune would be fully finished, I might make it into a library instead,
I'm still going to do that XACT thing in another PR.

PR #176, opened 2024-06-20
Enemies will no longer be removed automatically from battle when spared if exit on defeat is false

FireRainV:
@NyakoFox , check if this is okay.


jmxamongusmodder:
this. i like.

FireRainV:
This is now part of the key items PR.

PR #187, opened 2024-07-03
Revamped inventories conversion

FireRainV:
The dark inventory is no longer attached to the ball of junk, it's now a variable.
The light inventory is no longer a storage in the dark inventory, it's now a variable.

### DIFF
--- a/data/items/light/ball_of_junk.lua
+++ b/data/items/light/ball_of_junk.lua
@@ -1,6 +1,6 @@
 local item, super = Class(Item, "light/ball_of_junk")
 
-function item:init(inventory)
+function item:init()
     super.init(self)
 
     -- Display name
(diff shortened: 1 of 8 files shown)


NyakoFox:
There's a conflict. Other than that, I'll let @SylviBlossom review this.

FireRainV:
fixed that @MrOinky and fixed the conflict.

NyakoFox:
There's conflicts, and also will this break old save files?

MrOinky:
> > There's conflicts, and also will this break old save files?
> 
> It doesn't say that it has a conflict to me. 

The conflict you can't see is likely happening because you fixed the original conflicts via a merge commit instead of a rebase, you can fix it by rebasing on the latest commit of `upstream/main` instead (this will cause the merge commit to disappear, you will probably need to fix the conflict again as part of the rebase)

FireRainV:
> > > There's conflicts, and also will this break old save files?
> > 
> > 
> > It doesn't say that it has a conflict to me.
> 
> The conflict you can't see is likely happening because you fixed the original conflicts via a merge commit instead of a rebase, you can fix it by rebasing on the latest commit of `upstream/main` instead (this will cause the merge commit to disappear, you will probably need to fix the conflict again as part of the rebase)

like this?

PR #188, opened 2024-07-03
Text effect PoC

Dobby233Liu:
`bold`/`b`: `[b:(true/false)]`, the implementation is a pain
`strike`/`s`: `[s:(true/false/<height>),(<offset>)]`
`underline`/`u`: `[u:(true/false/<height>),(<offset>)]`

### DIFF
--- a/src/engine/objects/text.lua
+++ b/src/engine/objects/text.lua
@@ -2,7 +2,7 @@
 ---@overload fun(...) : Text
 local Text, super = Class(Object)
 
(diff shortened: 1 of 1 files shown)


Dobby233Liu:
How the hell do i italic
Also [.](https://github.com/sysl-dev/SYSL-Text/blob/master/readme.md#standard-tags)

Dobby233Liu:
This was a bust I think

PR #190, opened 2024-07-03
You can now specify a Boolean as a button type

FireRainV:
This will create an empty space, used as a cosmetic feature.

oldUIPositions config will now center the buttons (remains unchanged with 5 buttons)


polypoyo:
I feel like nil would make more sense. Could just be me though.

FireRainV:
> I feel like nil would make more sense. Could just be me though.

You can't specify nil in a table. Nil can be used to delete the space.

PR #194, opened 2024-07-12
Call setFrame to update the texture in Sprite:setFrames

Dobby233Liu:
This makes sure the frame number is clamped,
in case a sprite changes its animation to one
with lesser frames than the previous one.


NyakoFox:
Your PR comment
Looks like it's formatted like
almost a haiku

PR #197, opened 2024-07-18
Unusable light items can no longer be used

FireRainV:
the USE button will do nothing and will be greyed out.


NyakoFox:
This is a design decision I'm not sure about, because LW items should always be usable. Disabling them is a holdover from DW items...

PR #202, opened 2024-08-03
Add Tracker formats

polypoyo:
Adds searching support for XM, ImpulseTracker, and twenty other formats supported by ModPlug (and therefore LÖVE).

* [LÖVE Wiki](https://love2d.org/wiki/Tutorial:Audio#Formats)
* [ModPlug README](http://sourceforge.net/p/modplug-xmms/git/ci/master/tree/libmodplug/README#l39)


NyakoFox:
I guess I don't see much point, but if LÖVE supports it, I don't see why _not_.

PR #206, opened 2024-08-13
Game annotations

MrOinky:
quick one that adds type annotations for all of Game - I tried doing some weird luadoc meta stuff here to define SaveData as a custom type which i'd like looked over just because i think its quite funky (it also displays awfully in the code editor) just incase there is a better way to do this than through an alias


MrOinky:
Made some improvements here - the `SaveData`s are now documented as meta classes rather than tables, which I think is significantly cleaner

PR #209, opened 2024-08-16
Added self.can_die to enemyBattler

FireRainV:
False by default.
If the enemy was defeated though violence, the enemy will play the fatal animation.

SnowGrave is unaffected by this option.

This is to make it way easier to use the death animation.


NyakoFox:
We had something like this before and removed it in favor of overrides. Generally, we want to move away from booleans for everything, NOT add more.

PR #215, opened 2024-08-17
Fixed a bug where a debug menu would be still opened when it shouldn't

FireRainV:


NyakoFox:
I don't exactly like the hardcoding here -- it'd be better if they checked the result of their should-exist function instead

FireRainV:
what do you mean?

PR #227, opened 2024-09-03
Merged Test Mods

FireRainV:


NyakoFox:
The dark transition does not need a map property. There's a callback which is supposed to be used to switch maps.

FireRainV:
> The dark transition does not need a map property. There's a callback which is supposed to be used to switch maps.

Didn't know, I reverted that change and used the callback for the map transition.
If you're okay with the test mods being merged, this branch is complete and ready to be merged.

PR #232, opened 2024-09-04
made kris's battle/intro animation not loop

Idunno1:
ralsei's does, also i need this for a callback


NyakoFox:
susies should loop, there's a frame where it loops back to the beginning in DR

PR #233, opened 2024-09-07
Created a Kristal Event for when the engine has encountered an error

FireRainV:
Allows you to execute stuff when getting an error. Good example is just executing your unload code.


NyakoFox:
Needs conflicts resolved.

Hyperfae:
I think a better way to handle this would be to not set `Mod = nil` in the main loop, allowing Kristal.returnToMenu to handle this.

NyakoFox:
That mainly depends on where we want the callback, then -- immediately on error versus when they try to return.

Though, `Kristal.returnToMenu` DOES unload everything, doesn't it? So maybe that IS the better place... yes, I think that'd be worth doing.

ghost:
it doesn't unload anything when you return to menu since the mods and libraries aren't loaded anymore, but if they still have global variables that suppose to be cleared by mod/lib:unload(), it won't clear them.

ghost:
fixed the conflict

PR #244, opened 2024-09-20
Added all the missing key items

FireRainV:
BrokenCake, Lancer, Rouxls Kaard and Starwalker.

Better Cards light item.

Some other important fixes regarding battles.


NyakoFox:
Thanks!

PR #245, opened 2024-09-20
Add dmg_sprites var to enemies

TFLTV:
so you can individually get each enemy's dmg_sprites
you can do stuff like this:
```lua
function enemy:update()
    super.update(self)
    for _, i in ipairs(self.dmg_sprites) do
        i.y = i.y + 3 -- slowly move down by 3 pixels (Cant you just use speed_y?)
        i:play(1, false, function(s) s:remove() end) -- freeze the animation.. dont do it like this. this was just for testing.
    end
end
```

### DIFF
--- a/src/engine/game/battle.lua
+++ b/src/engine/game/battle.lua
@@ -1128,9 +1128,8 @@ function Battle:processAction(action)
                 local relative_pos_x, relative_pos_y = enemy:getRelativePos(enemy.width/2, enemy.height/2)
                 dmg_sprite:setPosition(relative_pos_x + enemy.dmg_sprite_offset[1], relative_pos_y + enemy.dmg_sprite_offset[2])
                 dmg_sprite.layer = enemy.layer + 0.01
-                dmg_sprite.onRemove = function() Utils.removeFromTable(enemy.dmg_sprites, dmg_sprite) end -- Remove the dmg_sprite from the enemy's dmg_sprite table when its removed
-                dmg_sprite:play(1/15, false, function(s) s:remove() end)
                 table.insert(enemy.dmg_sprites,dmg_sprite)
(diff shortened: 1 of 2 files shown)


FireRainV:
I found a bug in your code, the Sprite objects that already disappeared won't remove themselves from the table.

I don't think someone would need to make the attack location/animation different for different enemies, you can already do this tricks as part of the animation itself.

FireRainV:
Actually, it might be useful. Just remember to fix that bug I mentioned.

TFLTV:
> I found a bug in your code, the Sprite objects that already disappeared won't remove themselves from the table.

Ah, didnt know that was a bug, just thought that the sprites would never be removed for some reason (I thought that if the Sprite object was removed, it would also remove itself from the table). Ill look into it

FireRainV:
You can check if they no longer exist by checking if they have a parent, like this:
```
if dmg_sprite.parent then
    print("I exist!")
else
    print("I no longer exist...")
end
```

So simply use a reversed loop to delete the ones which don't have a parent anymore from the table. (it needs to be reversed to not mess with the array.)
best place is in `EnemyBattler:update()`

Do this:
```
for i = #self.dmg_sprites, 1, -1 do
    if not sprite.parent then
        table.remove(self.dmg_sprites, i)
    end
end
```

TFLTV:
Oh, i already got a solution that might(?) be better

i just inserted this
`dmg_sprite.onRemove = function() Utils.removeFromTable(enemy.dmg_sprites, dmg_sprite) end`

FireRainV:
lol, I didn't even know this function existed. Just wasted myself my own time coding this solution...

NyakoFox:
Why not put `Utils.removeFromTable(enemy.dmg_sprites, dmg_sprite)` after `s:remove()` instead?

FireRainV:
probably better to do what Ally said, but you'll also need to put the `table.insert(enemy.dmg_sprites,dmg_sprite)` above the play function to prevent an issue with party battlers who don't have an attack animation, where it will get executed frame 1.

PR #249, opened 2024-09-22
Remove bullet offscreen fix

FireRainV:
The reason I'm combining width and height is to deal with screen corners.


FireRainV:
It's to also deal with rotation apparently.

PR #250, opened 2024-09-22
make ball of junk accurate (again)

MrOinky:
While looking at something else i noticed that the ball of junk had unaccuracied itself as its `getCheck()` was missing, this readds it for the same chapter 1 accuracy as before

### DIFF
--- a/data/items/light/ball_of_junk.lua
+++ b/data/items/light/ball_of_junk.lua
@@ -86,4 +86,13 @@ function Item:onCheck()
     end)
 end
 
+function item:getCheck()
+    local check = self.check
+    if Game.chapter == 1 then
+        check = "A small ball\nof accumulated things."
+    end
+
+    return check
+end
+
 return item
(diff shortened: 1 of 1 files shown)


FireRainV:
I moved it to the self.check line, no need for this

FireRainV:
you can close this PR I think, its in the initial:
`self.check = Game.chapter == 1 and "A small ball\nof accumulated things." or "A small ball\nof accumulated things in your\npocket."`

MrOinky:
Didn't notice that it had been moved there, however it should ideally be moved back to getCheck() because its a dynamic text, defining it in init() means it will only update if re-initialised which it isn't after the first time the ball is recreated (in this case, its really just a good practice thing because its unlikely someone will change the chapter value mid mod but i think its better to leave it the way it was regardless)

FireRainV:
this is how we did it for other stuff, like how Ralsei will look like, whether Susie will have bangs.
The config menu won't reload if you change the chapter mid-game anyway, you're not suppose to modify the Game.chapter without reloading the mod.

PR #253, opened 2024-09-23
Added support for a table in self.check, allowing for multiple text boxes

FireRainV:
Also, made the ball of junk check better, no need to overwrite it with getCheck().


NyakoFox:
please don't revert the get check change

FireRainV:
> please don't revert the get check change

Ok, I will re-add it.

FireRainV:
@NyakoFox done, I re-added it.
But, we may need to update all the items, actors and party members to be dynamic like that if that's how you want it.

PR #254, opened 2024-09-23
Add `party` variable to `Character`s

MrOinky:
Adds the `party` variable to all Character objects, storing the ID of the associated party member to the Character if they were created from one, as the previous system of relying on actor IDs to connect them runs into issues if two party members make use of the same actor


FireRainV:
Love this addition!

PR #260, opened 2024-10-02
Move main script into `scripts/main.lua`, and change json config to lua config 

MrOinky:
This PR changes the system by which mods and libraries define their main script and config files, opting to move `mod.lua` and `lib.lua` into `scripts/main.lua`, while converting the config formats from JSON to lua, making all scripts now reside inside the *scripts* folder, and getting rid of the only json file that modders had to directly interact with (and all of JSONs evil quirks, you can now enjoy trailing commas and comments that don't cause a slew of angry errors). The contents of both files remain the same, except for configs being in a lua format that returns the config table instead of JSON.
New mods will all be created using this format.
Also fixed the thing where you can't escape the mod load error menu because it had no keybinds

## Breaking Changes
There are actually... none! Kristal continues to have legacy support for the old format with `.json` files (for now) with an `{}` indicator next to their version tag to indicate as such (nothing has actually changed that warrants making every old mod spontaneously combust Kristal)

## How to Port A Mod Tutoriel 101
This format is very simple to change to! You can do it in a few simple steps for both mods and libraries:
- Rename the `mod.lua` or `lib.lua` file, if it exists, to `main.lua`, and move it into the `scripts` folder
- Change the file extension of your mod config, `mod.json` or `lib.json`, to `.lua`
- Add `return` before the opening `{` on the first line of the config file
- Using your code editor's Find and Replace tool on the config, do all of the following:
- - Replace all instances of `[` with `{`
- - Replace all instances of `]` with `}`
- - Replace all instances of `//` with `--`
- - Replace all instances of `":` with ` =`
- Finally, remove the opening `"` for every key name (make sure you don't remove any from the ACTUAL strings by accident)


NyakoFox:
LGTM, ty!

PR #263, opened 2024-10-05
`mod.lua`, `lib.lua` config files and `scripts/main.lua` main scripts (attempt 2)

MrOinky:
I've come up with a new fix for the legacy checking which, while it appears a bit finnicky at first (reading the `lib.lua` file as text in the ambiguous case), after properly testing it, including on Dark Place this time (because it has a ton of libraries in particular, where the errors were arising) works properly, and as far as i can tell will not run into any of the previous problems
And to clarify further what the legacy checker is doing when it is reading `lib.lua` now, it is checking for a `return {` at the start, as this is how files on the *new system only* end up looking - it does not affect the format of anything still on the old system, and is only used as a last resort when there is no other indication of whether a library is using the legacy or new system, not the default way of checking


MrOinky:
Disregard my prior comments about reading the file in text as i have gotten rid of that awful 🤮 system now in place of executing the file again but much better than the time before with several safe checks on how it executes and the return values that i am fairly confident should now catch every library correctly (and never error in the loadthread)
For testing this time i used the same mods (most prominently Dark Place and all the libraries to go with it) and everything loaded fine - as an additional test i also threw every library through the new check even when the case wasn't ambiguous and nothing got incorrectly marked as the wrong thing so all in all i'm pretty happy that this solution accounts for everything - it should be 100% backwards compatible and not hindering to the new system in any way either

PR #264, opened 2024-10-06
actor minifaces and miniface text command

HUECYCLES:
notes: has a tiny bug regarding the first line of the text being pushed over to the edge a bit, and isn't compatible with portraits.
added a cutscene to testmod as well.


HUECYCLES:
i forgot to add the sweet miniface to the testmod lol

TFLTV:
couldnt this be done with `[image:mini_face]` ..?

HUECYCLES:
no, minifaces have indents with the text and animate like talk sprites.

TFLTV:
Ah ok

NyakoFox:
I don't see a check for dry runs anywhere, does the sprite still get spawned when you do a dry run?

HUECYCLES:
oh, what exactly is a dry run?

NyakoFox:
A dry run is where the engine goes through text nodes without modifying any lasting state outside of text state; dry runs are used for things like calculating the width and height of text, and automatic newline insertion. Meaning, things like `[sprite]` shouldn't actually create a sprite in a dry run -- we're not actually trying to *display* the text, just figuring out the size of everything.

HUECYCLES:
ah i see, so where exactly should i do a dry run check and how should i do it? sorry, i just based it off how portraits were done lol

HUECYCLES:
ok, checked the dry run and also added an extra argument for speed control.
hope this is fine! lemme know if there's anything else to do

HUECYCLES:
hey, it's been a while and i'm wondering if there's anything else i should do?

HUECYCLES:
ok fixed a stupid error actually lol

NyakoFox:
Are there any other issues with this other than the two mentioned in the OP? Is the OP still accurate as well?

HUECYCLES:
to my knowledge, no i don't think so? it might be good to go but it would be nice if someone could take a look at this and see if it has any other issues that i can fix in order for it to be merged

NyakoFox:
LGTM!

PR #266, opened 2024-10-07
[FIX] the window freeze when open mods folder on win11

Ferpakenameyea:
Use `os.execute` to open the mods directory.
If it's not on windows the code will fallback to the original plan.

This is because when I play on windows11, the game stucks **mostly** when I use the 'open mods folder' and it's annoying.

### DIFF
--- a/src/engine/menu/mainmenucredits.lua
+++ b/src/engine/menu/mainmenucredits.lua
@@ -57,7 +57,7 @@ function MainMenuCredits:init(menu)
                 "WIL-TZY",
                 "TFLTV",
                 "J.A.R.U.",
-                "",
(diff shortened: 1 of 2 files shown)


NyakoFox:
This seems to be a love2d issue, I've seen it happen with other projects. Could you add a FIXME comment?

Ferpakenameyea:
I've done adding this, don't know if I'm doing right?

NyakoFox:
Don't forget to add yourself to the credits!

Ferpakenameyea:
I've added it to mainmenucredits.lua

NyakoFox:
Thanks!

PR #267, opened 2024-10-07
Fix input bug in menu credits

TFLTV:
fixes this

https://github.com/user-attachments/assets/c95f5e4f-1d27-4380-a78a-71c3b3669093


TFLTV:
happens in Deltarune. + this would have to be a global fix

PR #268, opened 2024-10-07
Add pages to the keybinds menu for each mod's keybinds

MrOinky:
Aside from what is mentioned in the title, there are a few additional miscellaneous tweaks in this pull:
- `hideKeybinds` mod config to optionally hide all your mod's keybinds from the Main Menu (for mod authors who want to make thier own in-mod rebind system)
- New keybind `mod_rebind` - Used to open the Mod Controls menu (see below)
- Mod Controls menu - for mods that define keybinds, the `mod_rebind` key (default `/` on keyboard) can be pressed on the ModList screen to open up their specific keybind page (Will auto-switch between gamepad/keyboard if you change inputs!)
- `Input.mod_keybinds` table - okay this isn't a real feature but if for some reason you needed to know what keybinds a mod uses this table has a list of them (only includes the *keybind name* and *not* the bound keys)
- `Input.resetBinds(gamepad, mod_id)` - similar to above this is mainly for the controls menu but if you wanted to reset a specific mod's keybinds you can now do it with `resetBinds` here - or use "KRISTAL" as the `mod_id` to reset all of the kristal keybinds

An extra important clarification too: this does NOT change anything about how keybinds are stored, they continue to be shared between mods that define the same bind and will appear in all of those mods keybinds with the same bound key and update accordingly as it is changed

### DIFF
--- a/src/engine/input.lua
+++ b/src/engine/input.lua
@@ -77,7 +77,7 @@ Input.gamepad_cursor_x = (love.graphics.getWidth()  / 2) - (Input.gamepad_cursor
 Input.gamepad_cursor_y = (love.graphics.getHeight() / 2) - (Input.gamepad_cursor_size / 2)
 
 Input.order = {
-    "down", "right", "up", "left", "confirm", "cancel", "menu", "console", "debug_menu", "object_selector", "fast_forward"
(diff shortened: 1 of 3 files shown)


NyakoFox:
LGTM, thanks!

PR #269, opened 2024-10-08
add mouse input functions

vitellaryjr:
- Input.mousePressed(button), Input.mouseHeld(button), and Input.mouseReleased(button) have all been added to Input, returning whether the input is being done, the X and Y that the input is being done on, and how many presses are associated with the input (eg. for checking if the player double-clicked, i believe)
- Object:clicked(button) and Collider:clicked(button) have been added, as shorthands for checking whether the object or collider has been clicked by a specific button, without needing to manually convert screen-to-local yourself
- some stuff added to testmod for testing. rip ralsei

### DIFF
--- a/src/engine/input.lua
+++ b/src/engine/input.lua
@@ -78,21 +78,9 @@ Input.gamepad_cursor_x = (love.graphics.getWidth()  / 2) - (Input.gamepad_cursor
 Input.gamepad_cursor_y = (love.graphics.getHeight() / 2) - (Input.gamepad_cursor_size / 2)
 
 Input.mouse_button_max = 3
-Input.mouse_down = {
-    [1] = {x = 0, y = 0, presses = 0},
-    [2] = {x = 0, y = 0, presses = 0},
-    [3] = {x = 0, y = 0, presses = 0},
-}
-Input.mouse_pressed = {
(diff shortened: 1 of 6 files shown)


src/engine/input.lua
```diff
@@ -72,6 +77,23 @@ Input.gamepad_cursor_size = 10
 Input.gamepad_cursor_x = (love.graphics.getWidth()  / 2) - (Input.gamepad_cursor_size / 2)
 Input.gamepad_cursor_y = (love.graphics.getHeight() / 2) - (Input.gamepad_cursor_size / 2)
 
+Input.mouse_button_max = 3
+Input.mouse_down = {
+    [1] = {0, 0, 0}, -- x, y, presses
```
NyakoFox, line 82: please use keys instead of a comment, also you can have more than three mouse buttons so maybe you can assign the defaults elsewhere

vitellaryjr: as far as i know, there's no way to detect ahead of time how many mouse buttons the user might have; i adjust the default in Input.onMousePressed if the user ever inputs a mouse button higher than the current default

vitellaryjr:
don't merge yet whoops, need to make a change to mouseDown

PR #273, opened 2024-10-17
Tables for encounter text

TFLTV:
Automatically progresses through the table of the text. Add [wait] at the end for it to show for a longer time.

### DIFF
--- a/src/engine/game/battle.lua
+++ b/src/engine/game/battle.lua
@@ -161,6 +161,8 @@ function Battle:init()
     self.defeated_enemies = {}
 
     self.seen_encounter_text = false
+    self.table_encounter_text = {}
+    self.table_encounter_text_index = 1
 
     self.waves = {}
     self.finished_waves = false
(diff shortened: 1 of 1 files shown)


TFLTV:
Btw dont merge this yet it has a bug

TFLTV:
fixed

polypoyo:
Would this make more sense as a third-party library?

A quick-and-very-dirty proof of concept of that is [here](https://github.com/polypoyo/spamton-ex-kristalmod/tree/main/libraries/MultiEncounterText). All I did was replace methods that this PR modifies (:init is the only properly hooked one)

TFLTV:
> Would this make more sense as a third-party library?
yea

src/engine/game/battle.lua
```diff
@@ -2450,6 +2458,14 @@ function Battle:update()
                 self:setState("DIALOGUEEND")
             end
         end
+    elseif self.state == "ACTIONSELECT" then
+        if self.table_encounter_text_index < #self.table_encounter_text and self.table_encounter_text ~= {} then
```
FireRainV, line 2462: I see a bug with your PR, you can't compare different tables with == or ~=, they'll always be different.
If you want to check if it's not an empty table, use `#self.table_encounter_text > 0`.
There's also `Utils.equal()` you can use to compare if the tables are the same.

Also, I don't think this feature is needed to be added to the base engine, it may work better as a library.

Hyperboid: Wasn't it already packaged as a library?

FireRainV: Well, they forgot to close this PR then.

PR #275, opened 2024-10-28
New method: Actor:onTextSound(...)

polypoyo:
Pretty much does what the title says. ~~The only caveat is that the `[voice]` command doesn't work when the actor's `:onTextSound` returns true. Existing actors are unaffected.~~


polypoyo:
`[voice]` is fixed! I think this is ready to merge now.

NyakoFox:
LGTM, thanks!

PR #278, opened 2024-10-31
Fix borders on ultrawide aspect ratios

polypoyo:
This does not fix aspect ratios less than 16:9. It does change the way they are broken, though.

Tested with built-in static borders, and Dark Place Legacy's dynamic borders.


FireRainV:
I forgot that love.graphics.translate was a thing

PR #281, opened 2024-11-07
Border class and registry

polypoyo:
Makes it possible to split up your custom borders. Tested with `glow` and `battle` borders in Dark Place Legacy.

### DIFF
--- a/mods/_testmod/scripts/borders/spinnysquare.lua
+++ b/mods/_testmod/scripts/borders/spinnysquare.lua
@@ -1,0 +1,41 @@
+---@class Border.spinnysquare: Border
+local SpinnySquare, super = Class(Border)
+
+function SpinnySquare:init()
+    super.init(self)
+end
+
+function SpinnySquare:drawRect(max_depth, white)
+    if max_depth < 0 then return end
+    love.graphics.push("all")
+    if white then
+        Draw.setColor(1,1,1)
+    else
+        Draw.setColor(0.2,0.2,0.2)
+    end
+    love.graphics.scale(
+        Utils.clampMap(
+            math.sin(Kristal.getTime()),
+            -1, 1, 0.6, 0.93
+        )
+    )
+    love.graphics.rotate(math.rad(20))
+    love.graphics.rectangle("fill", -1000*BORDER_SCALE, -1000*BORDER_SCALE, 1000*2*BORDER_SCALE, 1000*2*BORDER_SCALE)
+    self:drawRect(max_depth - 1, not white)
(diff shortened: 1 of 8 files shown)


polypoyo:
> Tested with `glow` and `battle` borders in Dark Place Legacy.

Should clarify that I mean I tested that those work without modification.

FireRainV:
Make sure to also change the borders debug menu to use the registry instead of manually searching.

polypoyo:
Thanks for bringing that to my attention! For now, it will check both the files and the registry.

FireRainV:
why didn't you remove the old code? you're basically adding the same borders twice.

FireRainV:
nvm, it's intentional

mods/_testmod/scripts/borders/spinnysquare.lua
```diff
@@ -0,0 +1,41 @@
...
+    )
+    love.graphics.rotate(math.rad(20))
+    love.graphics.rectangle("fill", -1000*BORDER_SCALE, -1000*BORDER_SCALE, 1000*2*BORDER_SCALE, 1000*2*BORDER_SCALE)
+    self:drawRect(max_depth - 1, not white)
+    love.graphics.pop()
+end
+
+function SpinnySquare:draw()
+    love.graphics.push("all")
+    love.graphics.translate(1920*0.5*BORDER_SCALE, 1080*0.5*BORDER_SCALE)
```
FireRainV, line 30: what does this line do? it won't have issues for monitors that are non-1080p, right? I have a 1440p monitor.

polypoyo: This is supposed to move the 0,0 from the top left of the 16:9 to the middle of the screen for the drawRect method. I don't know if it works for 1440p.

FireRainV: yea, I changed to this branch and it seems to work fine, but would it look on non-16:9 screens like mobiles and widescreens?:
image

polypoyo: On 21:9 it works fine. The `love.graphics.translate` from my ultrawide PR applies here.

NyakoFox:
Does this break backwards compatibility?

polypoyo:
> Does this break backwards compatibility?

AFAIK no. I've tested with Dark Place Legacy's animated borders (KRISTAL_EVENT.onBorderDraw), and those are fine. Also tested with image borders in StarRune

PR #285, opened 2024-11-17
Fixed Hotswapper (for me at least)

Simbel0:
Despite apparently working for the devs, the hotswapper never worked for me no matter what I did, and I'm gonna assume I am not the only one in this situation

This basically adds the source directory to Lua's PATH variable when starting Kristal (and add a little check to remove it when trying to get some info on it). This seems to be enough to have `package.searchpath(key, package.path)` work as expected.

### DIFF
--- a/src/hotswapper.lua
+++ b/src/hotswapper.lua
@@ -10,6 +10,8 @@ Hotswapper.files = {
     registry = {}
 }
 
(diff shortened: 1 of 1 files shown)


NyakoFox:
lgtm

PR #287, opened 2024-11-20
enemies_index fix

FireRainV:
Prevents an enemy dupe when colliding with a ChaserEnemy who doesn't have an encounter attached. In addition, prevents a crash on end.

### DIFF
--- a/src/engine/game/battle/encounter.lua
+++ b/src/engine/game/battle/encounter.lua
@@ -22,6 +22,11 @@
 local Encounter = Class()
 
 function Encounter:init()
+    -- Sets the encounter ID incase it was started without an encounter file
+    if not self.id then
+        self.id = "_local"
+    end
+    
     -- Text that will be displayed when the battle starts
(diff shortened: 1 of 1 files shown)


src/engine/game/battle/encounter.lua
```diff
@@ -22,6 +22,11 @@
 local Encounter = Class()
 
 function Encounter:init()
+    -- Sets the encounter ID incase it was started without an encounter file
+    if not self.id then
+        self.id = "_local"
```
NyakoFox, line 27: what does "_local" mean? is this just a placeholder string? whats wrong with nil?

Hyperboid: > what does "_local" mean? is this just a placeholder string? whats wrong with nil?

It fixes a nil-concat error when calling Encounter:setFlag.
Admittedly, it would be better to make Encounter:setFlag set a flag on the initiating ChaserEnemy if it doesn't have an ID (and DOES have a ChaserEnemy)

PR #299, opened 2024-12-26
Saving Cell Calls to the game file

FireRainV:
Also, added Game.world:removeCall(name) and adjusted the functionality of Game.world:replaceCall(name, index (now old_name), scene)

BREAKING CHANGE!!!

### DIFF
--- a/src/engine/game/world.lua
+++ b/src/engine/game/world.lua
@@ -313,11 +313,27 @@ function World:registerCall(name, scene)
 end
 
 --- Replaces a phone call in the Light World CELL menu with another
----@param name  string          The name of the call as it will show in the CELL menu
----@param index integer         The index of the call to replace
----@param scene string          The cutscene to play when the call is selected
-function World:replaceCall(name, index, scene)
-    self.calls[index] = {name, scene}
+---@param name     string          The name of the call as it will show in the CELL menu
+---@param old_name string          The name of the call to replace
+---@param scene    string          The cutscene to play when the call is selected
+function World:replaceCall(name, old_name, scene)
+    for i,call in ipairs(self.calls) do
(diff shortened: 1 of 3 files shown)


src/engine/game/world.lua
```diff
@@ -313,11 +313,32 @@ function World:registerCall(name, scene)
...
----@param index integer         The index of the call to replace
----@param scene string          The cutscene to play when the call is selected
-function World:replaceCall(name, index, scene)
-    self.calls[index] = {name, scene}
+---@param replace_name string          The name of the call to replace
+---@param name         string          The name of the call as it will show in the CELL menu
+---@param scene        string          The cutscene to play when the call is selected
+function World:replaceCall(replace_name, name, scene)
+    for i,call in ipairs(self.calls) do
+        if call[1] == old_name then
```
SylviBlossom, line 321:
```suggestion
        if call[1] == replace_name then
```
Incorrect variable name here

SylviBlossom (changes requested):
This looks like a good system to me, however I'd suggest that, if changing and breaking the existing cell system, it should be moved out of `World` into `Game` directly, since it's static data and has nothing to do with the temporary world state, especially given it's saved to the file now.

To fit with the addition of `removeCall`, also consider renaming `registerCall` to `addCall`.

src/engine/game/world.lua
FireRainV: fixed

FireRainV:
I've done the changes you requested

FireRainV:
@SylviBlossom The reason I kept it as `registerCall` instead of `addCall` is because we're also using `registerAct` while there is a `removeAct`.

PR #305, opened 2024-12-28
PartyMember.soul_offset getter

FireRainV:


FireRainV:
If you're asking why it is a getter, it's because it may be different if you're using a different battle sprites in another battle. Like in the Frisk library. (or Ralsei)

PR #308, opened 2024-12-31
The candy box can no longer kill

FireRainV:
It won't go below 1 HP.
I think it's stupid to die from it. You don't die from the bad memory either.


NyakoFox:
Generally we don't merge things because "I think it's stupid"... if people want this behavior in their mods then they can do it themselves, or maybe someone could even make a small library for it... either way, I don't think this fits in the engine, sorry.

ghost:
okay

PR #312, opened 2025-01-04
Configurable Susie style in the light world

FireRainV:
It wasn't configurable like her dark world style.


ghost:
I think this one can be merged, it's a simple clean fix.

PR #313, opened 2025-01-04
Auto heal inaccuracy fix

FireRainV:


FireRainV:
There were also issues when attempting to heal while having over the maximum health, which was also fixed.

ghost:
apparently it was already fixed

PR #320, opened 2025-01-09
fixed bugs with the engine

FireRainV:
Stuff like the soul color not resetting when hovering on "create mod". "Start Game" not letting you choose a name when having no saves. Kristal.hasAnySaves() checks for non save files (like global file) and more.

### DIFF
--- a/data/actors/lancer_keyitem.lua
+++ b/data/actors/lancer_keyitem.lua
@@ -1,7 +1,7 @@
 local actor, super = Class(Actor, "lancer_keyitem")
 
 function actor:init()
-    super:init(self)
+    super.init(self)
 
     -- Display name (optional)
     self.name = "Lancer"
--- a/data/party/susie.lua
+++ b/data/party/susie.lua
@@ -155,7 +155,7 @@ end
 function character:getGameOverMessage(main)
     return {
         "Come on,[wait:5]\nthat all you got!?",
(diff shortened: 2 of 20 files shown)


Hyperboid:
Commits 50f3d1cc4744ed6127c6ca2d407065261d9a9df1 through 7756dcee613b2bbc74e8d625479c1cd084818db9 should probably be squashed.

FireRainV:
> Commits [50f3d1c](https://github.com/KristalTeam/Kristal/commit/50f3d1cc4744ed6127c6ca2d407065261d9a9df1) through [7756dce](https://github.com/KristalTeam/Kristal/commit/7756dcee613b2bbc74e8d625479c1cd084818db9) should probably be squashed.

Didn't even know it was a thing. Done it.
Probably will also be useful in other PRs.

NyakoFox:
I'll try to sift through this commit history and apply these myself... thank you for fixing a lot of these, but I really cannot merge this in its current state. The history is a complete mess and squashing would make it worse.

Generally you do want to split different changes into different commits. Maybe even a different PR for each fix, so some fixes don't get held up by problems with other patches. And there's definitely a lot of commits here (including some merge and revert commits), which even contain more than one patch/fix per commit...

To save you some time fixing this, and some back-and-forth from me, I'll merge these in manually.

PR #322, opened 2025-01-12
Add slide animation to kris_lw actor

TFLTV:
i didnt test this because i dont have kristal at the moment but should work


FireRainV:
I just tested your PR and it works fine

PR #323, opened 2025-01-17
Fix the underscore issue for mirrors

AcousticJammYT:
https://discord.com/channels/899153719248191538/1329851397255991306


Hyperfae:
I guess ACJ decided there were too many useless merge commits?

AcousticJamm-dev:
so the issue is I need to remake this because all that time ago I forgot to put it on a branch, and you can see what happened due to that

PR #327, opened 2025-01-24
Fixed the issue with the frame not getting drawn in the light world

FireRainV:
Calls love.draw() before a game over so you'll actually see changes like the 0 HP or any special sprite.


ghost:
when this PR will be looked at? I need it to fix an issue with my "But it refused..." game over in MGR.

PR #335, opened 2025-02-04
Mod select grid mode

Hyperboid:
An alternative mode for the mod list that works better when many mods are installed.


Hyperboid:
Converted to draft because apparently having this enabled breaks TARGET_MOD.

NyakoFox:
I hate to close this PR, but this is a bunch of design decisions that weren't ran by us first -- I can get pretty intimidated by PRs, and something like a bunch of design does not help.

Aside from that, this looks like a *ton* of upkeep, and not really worth it for us to include when the current one works fine.

Maybe a plugin would be better?

PR #338, opened 2025-02-11
Deprecate `super:` syntax

Hyperboid:


NyakoFox:
I agree with this PR actually -- we've been talking about deprecating this for a really long time. LGTM.

PR #339, opened 2025-02-23
Add emergency restart bind to errorhandler

MrOinky:
For those times where you create an error loop that can't be fixed without a hard reset (i do this too often)
The bind for this is `CTRL+R` (i wanted `CTRL+Esc` but windows uses it...)

### DIFF
--- a/src/kristal.lua
+++ b/src/kristal.lua
@@ -847,6 +847,8 @@ function Kristal.errorHandler(msg)
                 else
                     return "reload"
                 end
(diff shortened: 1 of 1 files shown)


Hyperboid:
Ctrl+Alt+Shift+R may be preferable for consistency with the rest of the engine.

MrOinky:
i was going to mention that i think `Ctrl+R` makes sense because this is contained specifically to the errorhandler but suddenly i realised that `Ctrl+Alt+Shift+R` is already a Kristal bind for force restart and works in the errorhandler

Oops!

PR #343, opened 2025-03-09
Support for loading Tiled polygon shapes as Event collision

MrOinky:
This PR allows for event colliders to take the form of Tiled polygons in maps. Custom events should automatically support this feature, and most collision-based Kristal events will support polygons too. Events that do not support polygons will turn into rectangles ingame if placed as a polygon (rather than the previous single-pixel hitbox).

(This does not add polygon support for most graphical events e.g. mirrors, silhouettes, outlines, as they rely on canvases, not collision)

### This PR includes breaking changes:
- The `Event` constructor `Event(x, y, w, h)` has been removed. `Event(x, y, shape)` must be used instead.
- As well as changing the constructor for `Event`, all Kristal event constructors that used an `x, y, w, h, properties` format of constructor now use `x, y, shape, properties`.

The error `Attempt to index local "shape" (a number value)` will appear when loading an event impacted by this breaking change.

### Quick fixes for breaking changes:
- If you have overriden `Event` or any base kristal event that previously used width and height (`w, h,`) in it's `init`, replace them with `shape,`. 
- If you used the `w` or `h` of the above `init` functions in your own code they should be replaced with `shape[1]` and `shape[2]` respectively.
- Custom events (files in `scripts/world/events`) that used any of the changed or removed `x, y, w, h` format event constructors for `super.init()` should switch to `x, y, shape` - `shape` is `w` and `h` as a single table, plus `data.polygon` as a third element for polygon support (e.g. `{w, h, data.polygon}`). 

(In short: turn widths and heights into the `shape` table, which is `{width, height, polygon}` if you have to create it manually. `polygon` data comes from `data.polygon` in this case).

### Non-breaking changes:
- Events can now load with polygon shaped collision from map data. 
- The `Event(x, y, size)` constructor is now `Event(x, y, shape)`. Previous usage of this constructor works as usual, but `shape` may also contain polygon data to read in the third index.
- Events which cannot use polygons now use a rectangle that encases the entire area of the polygon rather than a single pixel.

### DIFF
--- a/mods/_testmod/scripts/world/events/banana.lua
+++ b/mods/_testmod/scripts/world/events/banana.lua
@@ -1,7 +1,7 @@
 local Banana, super = Class(Event)
 
 function Banana:init(data)
-    super.init(self, data.center_x, data.center_y, data.width, data.height)
+    super.init(self, data.center_x, data.center_y, {data.width, data.height})
 
     self:setOrigin(0.5, 0.5)
     self:setSprite("banana", 0.25)
--- a/mods/_testmod/scripts/world/events/darkdoor.lua
+++ b/mods/_testmod/scripts/world/events/darkdoor.lua
@@ -1,7 +1,7 @@
 local DarkDoor, super = Class(Event)
 
 function DarkDoor:init(data)
-    super.init(self, data.center_x, data.center_y, data.width, data.height)
+    super.init(self, data.center_x, data.center_y, {data.width, data.height})
 
     self:setOrigin(0.5, 0.5)
     self:setSprite("darkdoor_closed")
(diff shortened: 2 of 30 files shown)


NyakoFox:
This seems like a rather risky breaking change -- although maybe that's fine?

What's `polygon` in the shape table if it's a rectangle? I worry about this change slowing things down if rectangles are now polygons -- as rectangles are normally much faster to check for collision against.

MrOinky:
This should be all good on that front - `polygon` is `nil` if the shape wasn't explicitly made via the polygon tool in Tiled and therefore `PolygonCollider` is only used in those cases - `Hitbox` will still be used as the collider for everything that is a rectangle (unless you draw your rectangles with the polygon tool...)

NyakoFox:
LGTM, thanks!

PR #346, opened 2025-04-06
Unhardcode ActorSprite aura color

Agent517:
Although it still defaults to red, aura colors can now be changed with the `aura_color` variable. Additionally, enemy events can be given an `aura_color` property, as demonstrated here by The Seven Ralsei Dummies.
image

### DIFF
--- a/src/engine/game/common/actorsprite.lua
+++ b/src/engine/game/common/actorsprite.lua
@@ -26,6 +26,7 @@
 ---
 ---@field aura              boolean                             Whether the sprite currently has a glowing aura (used for `ChaserEnemy` objects)
 ---@field aura_siner        number                              A timer used for the aura effect
+---@field aura_color        table                               The color used for the aura effect
 ---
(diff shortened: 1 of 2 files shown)


NyakoFox:
I do wonder if this'd be better as a library that does a few hooks... it being a variable in `ActorSprite` is acceptable but adding it as a property feels weird as something to maintain within base Kristal?

Hyperboid:
Good point. Although, it would be wise to split aura-drawing into a separate method so that such a hook wouldn't be ridiculously long. Or, and this is a bit of a stretch, make it a DrawFX?

NyakoFox:
I think a separate method would be best, although the variable in `ActorSprite` does work for now -- my only concern is the property part, not the change to extract the colors and make it a color table.

Agent517:
If the only pressing issue with this PR is the property for enemies, then I'll be removing it for now.

PR #349, opened 2025-04-12
Release mode prototype

Dobby233Liu:
This is the idea I mentioned on Discord 2 months ago https://discord.com/channels/899153719248191538/900166953598058559/1342071541436780575

### DIFF
--- a/src/engine/game/debugsystem.lua
+++ b/src/engine/game/debugsystem.lua
@@ -1378,6 +1378,10 @@ function DebugSystem:onWheelMoved(x, y)
 end
 
 function DebugSystem:draw()
+    if self.state == "IDLE" and self.menu_anim_timer >= 1 then
+        return
+    end
+
     love.graphics.setFont(self.font)
     Draw.setColor(1, 1, 1, 1)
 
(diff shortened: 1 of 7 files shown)


NyakoFox:
This looks good, but I'm worried about the design -- I did want to lock a bunch of debug things behind a "dev mode" but you'd still have the ability to turn debug *on* at any moment, to make it easier to debug when a user has a crash -- you'd be able to tell them "hey turn on debug and see if it still happens if you enter this in the console" or something... maybe I'm overcomplicating things, though. Either way, maybe a flag to forcibly turn back on debug would be good?

Something I've wanted to do for a while was a "dev mode" in mods, which would be for creating them, so debug mode and stuff would be locked behind it (with, again, the ability to reenable debug for released stuff if required), and you'd just turn it off when your mod is done.

So honestly, I'm not sure about this -- I'll need to think about this more.

Hyperboid:
If we _do_ end up going through with this, I think it would be a good idea to pre-load mod assets on startup, like https://github.com/Hyperboid/Kristal-And-DPR/commit/0d01129f33c88ca90430e8e24c0b933b6dcddfda. Note that this implementation is flawed as for some reason I decided it would be a good idea to make the main menu show up before loading finishes?

NyakoFox:
Mostly obsoleted by ccc654b36e7f438933a4cf2a541bea241ca01876. Taking a quick look through the changes, most seem obsoleted by the commit except for the draw optimization with the debug menu. I'll manually merge that one and give you commit credit.

PR #350, opened 2025-04-18
Fix Utils.absoluteToLocalPath returning nil when the file is in a lib

Hyperboid:
As branch name implies, this allows using library tilesets in mod maps.


NyakoFox:
Can I ask why this is in a block? The only reason for a loose block I can think of is variable scoping which doesn't seem much of a problem in this case.

PR #352, opened 2025-04-27
Utils.hook() but cool

Hyperboid:
Basically lets you use Utils.hook with the benefits of LSP completions.

### DIFF
--- a/mods/_testmod/scripts/hooks/Object.lua
+++ b/mods/_testmod/scripts/hooks/Object.lua
@@ -1,0 +1,22 @@
+---@class Object : Object
+local Object, super = Utils.hookScript(Object)
+
+function Object:preparefunny()
+    if self.funny_rng then return end
+    local rng = love.math.newRandomGenerator((tonumber((tostring(self)):sub(18), 16)/10000))
+    for i=1,6 do
+        rng:random()
+    end
+    self.funny_rng = (rng:random()-0.5) * 2
+end
+
+function Object:applyTransformTo(transform, ...)
+    self:preparefunny()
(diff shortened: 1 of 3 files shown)


NyakoFox:
Is there any way to make this work with current hooks? We don't really want duplicate systems -- honestly if this is similar enough we can deprecate old hooks for it...

(Ideally we'd have mixins already which would hopefully work with LSP but that's taking a while... so improving hooks sounds nice)

Hyperboid:
Syntax is basically identical to typical scripts/hooks files, so yeah, we can deprecate that.

NyakoFox:
Could we then replace `Utils.hook` with this to avoid breaking old hooks (and not introduce another system on top of hooks which are already two systems, being file based and function based)? Or are there edge-cases to worry about?

Hyperboid:
Not sure I follow. This is a sort of combination of the two systems, file hooks that work like Utils.hook under the hood.

...I'm starting to realize the possibility of an XKCD 927 situation.

NyakoFox:
Oh sorry, I understand now -- I was thinking this was something different (looked at it a few days ago, forgot about it, then took another look today assuming it was something it isn't)

This definitely feels a lot nicer than old file hooking... since it's a replacement of the old system, we should deprecate that (with a log warning on mod load probably?) and later on, prevent old file hooking.

I do want to talk to @SylviBlossom about this before merging it (as she knows most about the implementation of classes and hooks and files etc.) so I'll do that soon, hopefully tomorrow if I don't forget.

NyakoFox:
We should probably still deprecate the old system, but this looks good! Thanks!

PR #359, opened 2025-05-30
Add ability to have "thin" layers that don't contribute to map depth

Hyperboid:
Useful if you have a lot of conditional NPCs/objects in a single room. Basically, create a Group Layer with the property `thin` set to true, and add object layers within. This time I'm not sending it to Dark Place on accident.


NyakoFox:
LGTM, thanks!

PR #362, opened 2025-06-12
TIRED/Pacify spells changes for Ch 3/4

MrOinky:
Adds TIRED messages, AWAKE messages, and the flashing effect on enemy names and pacifying spells when the Chapter is set to 3 or 4 (or their respective configuration options are enabled).
(Also adds Chapter 3 and 4 config files + lets them be selected in mod creation)

My implementation for the spell changes specifically may be slightly suboptimal - while it's got no immediate problems, using color as a function feels off but otherwise like the most appropriate way to allow dynamic colouring
In that regard, there is also a **breaking change** for if the `color` component of a battle menu item is accessed after being added (with `Battle:addMenuItem()`). Now you get the color by calling it as a function e.g. `item:color()` rather than indexing it like `item.color`


MrOinky:
Closing this to split it into two (Chapter Configs and TIRED changes) because they're both in one commit here and that's Bad and also will probably explode into merge conflict hell as more chapter 3/4 PRs come in (and i can't fix that with this commit history, probably)

PR #363, opened 2025-06-12
lightworld save point+menu behavior

Agent517:
Made the following changes to save points in the Light World, based on Chapter 4:
- Save points will be partially transparent until approached
- The light save menu is now nearly identical to the dark one, only with storage and recruits removed. In its place is a "Return to Title" option

Unrelated to chapter 4, fixed a bug where playtime for saves did not show hours on the save list.


NyakoFox:
LGTM -- a few weird formatting things (ex. a bunch of newlines) but I don't think they should block this PR.

PR #365, opened 2025-06-14
attack effect overrides for weapons

Agent517:
The `attack_sprite`, `_sound`, and `_pitch` variables can be used to give unique attack effects to weapons, which take priority over those of the party member using them.

There is a weapon in chapter 3 that could make use of this, although I haven't added it, and won't say what it is due to potential spoilers.

### DIFF
--- a/data_templates/items/template.lua
+++ b/data_templates/items/template.lua
@@ -41,6 +41,13 @@ function item:init()
     self.bonus_name = nil
     self.bonus_icon = nil
 
+    -- Effect shown above enemy after attacking it with this item
+    self.attack_sprite = nil
+    -- Sound played when attacking
+    self.attack_sound = nil
+    -- Pitch of the attack sound
+    self.attack_pitch = nil
+
     -- Equippable characters (default true for armors, false for weapons)
     self.can_equip = {}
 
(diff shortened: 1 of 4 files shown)


NyakoFox:
As these are rarely used, these might be better as overrides which default to, well, the defaults (ex. it'd be implemented as `laz_c` and weapons can override the functions if needed).

The variables are nice but I worry they clutter the files too much -- things that are rarely changed should be overrides instead IMO.

Other than that, LGTM!

Agent517:
In that case, would it be alright to add some arguments to these while I'm at it? I was thinking the `battler` using the weapon, the `enemy` being attacked, and your `points` for timing the attack could all be fairly useful.

As for the defaults, I think it may be better to have those all be `nil`, so the party member's ultimately get used instead.

NyakoFox:
Arguments sound really good for that, yeah!

I forgot the battlers have a system for this already -- I wonder what should be prioritized, the weapon or the battler?

Maybe the weapon could default to returning what the battler does (prioritizing a weapon override basically)

So your weapons, by default, would just call the getter in the battler instead, and return that.

Agent517:
That works for me! I've made those changes with my latest commit.

NyakoFox:
I'll review again in the morning -- The playSound in the sprite getter is a little strange (we should have an on-weapon-use sort of callback honestly) but I guess it's the testmod so it's probably fine to merge -- I've been meaning to clean that mod up anyway

NyakoFox:
LGTM, thanks!

PR #366, opened 2025-06-14
Chapter 3/4 Config files + Basic Chapter changes

MrOinky:
Since people will be PRing Chapter 3+4 changes i figure that having config files in place for both of these will be important!
Also adds updated Stats, Titles, Powers, and Spell Descriptions from the new Chapters for when these chapters are chosen in mod create (or set by any other means)
Susie's heal spell and it's related power stat are unchanged as i've only been working through non-mechanical changes for this PR and those are both tied to new mechanics of that spell


NyakoFox:
LGTM, thank you!

PR #367, opened 2025-06-14
TIRED/Pacify changes for Chapter 3+4

MrOinky:
(Redo of #362, This version of the PR is based on #366 so that Chapter configs can be committed seperately in case this has to be reverted or something)

Adds TIRED messages, AWAKE messages, and the flashing effect on enemy names and pacifying spells when the Chapter is set to 3 or 4 (or their respective configuration options are enabled).

My implementation for the spell changes specifically may be slightly suboptimal - while it's got no immediate problems, using color as a function feels off but otherwise like the most appropriate way to allow dynamic colouring
In that regard, there is also a **breaking change** for if the `color` component of a battle menu item is accessed after being added (with `Battle:addMenuItem()`). Now you get the color by calling it as a function e.g. `item:color()` rather than indexing it like `item.color`


NyakoFox:
As the depending PR got merged, somehow the commits had different IDs -- you might want to rebase on top of master and drop the duplicated commits as they're causing merge conflicts. If you use GitHub Desktop, I apologize

MrOinky:
Rebased and the conflicts seem to have cleared up on my end! Let me know if the git history is still messed up in some way i've not noticed

NyakoFox:
LGTM -- there's a few changes I'd make, but unfortunately those are out of the scope of this PR (& would require a lot of battle menu changes to fix, so...)

PR #374, opened 2025-06-24
Config for the new magic glass

AcousticJammYT:
I just feel like the new magic glass sprite should be used if it's in the engine. Let me know if any of the chapters' configs is wrong.

### DIFF
--- a/configs/chapter1.json
+++ b/configs/chapter1.json
@@ -42,5 +42,7 @@
     "pushBlockInputLock": true,      // Whether to lock the player's input when pushing a block
     "keepTensionAfterBattle": false, // Whether TP should be kept after battle instead of reset
     "overworldSpells": true,         // Whether spells should be usable in the overworld
-    "damageUnderflowFix": true       // When disabled, negative enemy damage heals the enemy and displays a negative damage number
+    "damageUnderflowFix": true,      // When disabled, negative enemy damage heals the enemy and displays a negative damage number
+	
+	"newMagicGlass": false           // Whether MagicGlass uses the new sprite or not
 }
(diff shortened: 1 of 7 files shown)


MrOinky:
Chapter 3 uses the old `spr_magicalGlass_classic` sprite in DELTARUNE, so only Chapter 4 should use the new one
As for the config name it might make more sense to be `oldMagicGlass` with the condition inverted for consistency with the names of other visual changes in the config?

AcousticJamm-dev:
Got it, will make the change

AcousticJamm-dev:
Both suggested changes have been made.

nightpool:
@AcousticJammYT I noticed that you're using tabs instead of spaces here. You can see in the Github web UI that this is leading to mismatched indentation / whitespace errors. I'm not sure what the maintainers' opinion is, but I would suggest updating the formatting and whitespace style to match the rest of the file:

image

Which editor did you use to edit this file? VSCode will read the `.editorconfig` file at the root of this repository and configure your tab/space settings automatically, among other formatting options.

AcousticJamm-dev:
I hate how Github doesn't automatically convert them. And I'm actually using N++.

In any case, fixed.

NyakoFox:
This has been left open for so long because quite honestly I don't really get the point? DR is also incredibly inconsistent with the magical glass sprites it uses.

AcousticJamm-dev:
In that case, would you be more likely to accept it if I made it a property of the event instead, or should I just close this PR?

NyakoFox:
Feel free to make it a property, yeah.

AcousticJamm-dev:
Change made, it is only a property now

PR #378, opened 2025-07-02
Spell select and cast animations

CosmicPikachu001:
Spells would have the new fields `select_anim` and `cast_anim`, allowing for different spells to have different animations.

Also fixes a typo in `src/engine/game/common/data/spell` where it says 'disaply' instead of 'display'.


NyakoFox:
Oh this looks great, thank you!

PR #379, opened 2025-07-08
Fix sparing without an animation breaking battles

MrOinky:
Light World Kris and Susie will no longer cause the fabric of battles to break entirely as they inanimately stare you down... (Or any party lacking spare animations)


NyakoFox:
Are they stupid why do they do that

PR #380, opened 2025-07-16
Load all texture formats that LOVE supports

Hyperboid:


NyakoFox:
This one's been sitting here for a while solely because I don't really want to say this -- but if anything, we should *remove* `jpg` rather than add new formats. We don't want to encourage things like jpg so it shouldn't really be allowed by default... and honestly, I don't know anything about the other formats.

Is there an argument for why we SHOULD accept these?

ghost:
if someone is using .jpg, they probably already know what they're doing.
It would probably will be a big meme image, where lossy compression won't matter as much.

It's nice to allow more option for people without forcing them to convert these images to PNGs.

Hyperfae:
There isn't any real argument for this, no. At the time, I was trying to load an OpenEXR image to use as a high-quality normal map. Looking back on it, for the rare cases where an unusual image format is preferred, `love.graphics.newImage` will do just fine.

PR #387, opened 2025-07-27
Added :addTemporaryMercy function to EnemyBattler

HmmNoPls:
This is used by some enemies such as Shadowguy and Organikk in Deltarune.
DamageNumber has been modified to make changing the current text possible
Additionally, fixed a small typo in hotswapper.lua

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .vscode/
+.DS_Store
 
 mods/*
 !mods/_light/


NyakoFox:
A cleaner commit history would be nice next time, but I can squash this time. Thanks!

(Also, never make a PR from your main branch, because each PR needs to be its own branch.)

PR #391, opened 2025-08-03
Element based damage reduction

mario64thane:
Add element variables to items and bullets for element based damage reduction

### DIFF
--- a/data/items/armors/dealmaker.lua
+++ b/data/items/armors/dealmaker.lua
@@ -50,6 +50,11 @@ function item:init()
         ralsei = "Two pairs of glasses?",
         noelle = "(Seems... familiar?)",
     }
+
+    -- This item's element
+    self.element = "puppet/cat"
+    -- The multiplier for damage reduction based on this element
+    self.element_reduce_amount = 0.40
 end
 
 function item:applyMoneyBonus(gold)
--- a/data/items/armors/frayedbowtie.lua
+++ b/data/items/armors/frayedbowtie.lua
@@ -53,6 +53,11 @@ function item:init()
         ralsei = "It's still wearable!",
         noelle = "(Reminds me of Asgore...)",
     }
+
+    -- This item's element
+    self.element = "puppet/cat"
+    -- The multiplier for damage reduction based on this element
+    self.element_reduce_amount = 0.15
 end
 
 function item:getShopDescription()
--- a/data/items/armors/mannequin.lua
+++ b/data/items/armors/mannequin.lua
@@ -49,6 +49,11 @@ function item:init()
         ralsei = "Um, the d-dress is cute...",
         noelle = "(Why did they spend $300 on this!?)",
     }
+
+    -- This item's element
+    self.element = "puppet/cat"
+    -- The multiplier for damage reduction based on this element
+    self.element_reduce_amount = 0.35
 end
 
 function item:canEquip(character, slot_type, slot_index)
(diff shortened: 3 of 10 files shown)


NyakoFox:
In Kristal, we'd want elements to be string based rather than "magic number" IDs. Could you use string IDs instead?

(In the future, before making a PR for Chapters 3 & 4, some discussion with the maintainers would be nice, as we have our own plans for certain things.)

mario64thane:
Sorry for the late response, I assume this is what you mean by string IDs?

NyakoFox:
Looks good to me, being honest I don't know much about elemental damage, so I'll have to wait for other people to report issues and stuff once I merge this.

You've missed a few elements in the item code however -- the DealMaker and the Mannequin also have elemental reduction (cat/puppet I think?)

MrOinky:
I think the element variables in items and bullets should ideally have equivalent getter functions used to access their element information (as other data values do) rather than reading directly from the variable

mario64thane:
I have added the missing elements based on [this list](https://deltarune.fandom.com/wiki/Elements#:~:text=from%20elemental%20resistance.-,Armors%20with%20Elemental%20Resistance,-Resistance%20IDs%20and)

MrOinky:
Saw that you added getters, but currently the code is still using the variables directly - they'll need to be changed over too, otherwise the getters aren't actually being used
Also, while it shouldn't have a functional impact, `element` locals (e.g. in `PartyBattler:hurt()`) should probably default to `""` rather than `0` everywhere now that it's on a string ID system (if only to make the language server not complain about incorrect typing)

NyakoFox:
Recently I've been told that elements work a lot different than I initially thought -- have you been cross referencing how they work in DR? We might want to hold off on merging this until we're sure they work the way they do in DR.

Implementation parity does matter as we're not sure what future chapters will do with elements.

TheRektafire:
Aren't elements in actual deltarune literally just fake? Shortly after chapter 3 and 4 came out I recall seeing a tweet about some code for either tennas star attacks or one of the knight:s attacks  that just checks to see if you have the shadow mantle equipped and reduces the damage of the attack if so, there's no actual type effectiveness tables or lookups despite what the item descriptions in the game would imply, its literally just a bunch of attack specific hardcodes for specific items afaik, if anything adding an actual element system would be an improvement, at least from a modder standpoint if not an accuracy one

MrOinky:
elements in DELTARUNE have always been real, i'm not sure what you saw but it was either wrong or misrepresenting the system as since Chapter 2 even, `armorelementtemp` and `armorelementamounttemp` have existed and are matched up to the `element` values of applicable bullets (and then used in damage calculations), they're definitely not hardcoded

MrOinky:
Okay, I was checking something unrelated and ran into some code that actually IS hardcoded, so it's a bit more chaotic and un-coordinated, it looks like some fights and bullets specifically check for the shadowmantle (armor id 23) and apply special modifications to damage if it's present (sometimes to the bullets themselves, sometimes during damage calculation), but i've only found one example in the knight that actually also skips over regular elemental damage reduction when this case occurs, meaning most of them are also firing off the regular elemental damage system *plus* these hardcoded multipliers for the mantle (however, it looks like a lot of these also forgo actually setting elements on their bullets)

The system probably needs a better look at by someone else, but based on my very rough sifting through the code it looks like these two systems both handle elements and therefore implementing them accurately might need a bit more thorough consideration

FireRainV:
maybe turn the element from a simple string to a table of strings?

Like this:
from `self.element = "dark/star"` to `self.element = {"dark", "star"}`

NyakoFox:
no, strings are what we want

PR #402, opened 2025-08-25
move `FlashFade` logic to `draw`

MrOinky:
This should fix the occasional strange flicker effect that can happen when creating `FlashFade`s (most noticable when healing in battle)


NyakoFox:
We don't do logic in draw -- couldn't this be easily fixed by just setting `self.alpha` to 0 in init?

MrOinky:
Oh. yeah that would fix the issue too

PR #403, opened 2025-08-27
Begin Utils split -- NEEDS TESTING!

NyakoFox:
This is a massive PR, and definitely a symptom of "we'll do it later", knowing fully well that it gets worse and worse as time passes.

The amount of functions in `Utils` is insane, and I can't believe it's entirely documented (thank you, contributors!) but it's far too much. People understand black boxes more than Utils. Someone asks if a function exists, someone replies with "Yep, it's in Utils!" and they roll their eyes so far, they fall into their skull. Utils, a small container class for various common utilities, has outgrown its purpose and become a monster that nobody can tame. It's not a small container, it makes up 90% of engine code. I made that statistic up on the spot, but I bet you believed it anyway. That's how bad this is.

This commit tries to split Utils. Keyword: TRIES. But it's so big, so massive, that it's intertwined in everything, like a tree planting its roots. It's so deep in, you can find those roots anywhere you dig. So, this commit has to not only split Utils, but make the Utils functions redirect to their split versions, to not horribly break EVERY SINGLE MOD IN EXISTENCE. If that's not bad enough, I did a quick search -- and the pattern `(?<!\w)Utils\.` matches 810 times inside Kristal itself! And that's after I replaced a ton of them!

Ranting aside... it's really bad. Not every function was moved out of Utils unfortunately, and it seems like I'd be here for days if I were to manually replace every single instance of `Utils` in the engine.

We'll probably keep the Utils module around for 0.10.0's release, and start trying to remove it during 0.11.0's development cycle.

Either way... the point of this commit is to organize everything, while NOT breaking every single mod. Aside from a few behavior changes, most things should still work perfectly. If not, please report this to us.

Known behavior changes:
- `Utils.numberFromKey(t, name)` has been removed
- `Utils.keyFromNumber(t, number)` has been removed
- `Utils.hexToRgb(hex, value)` now supports hex codes which aren't just #RRGGBB, so `value` can now be a multiplier if you enter hex with alpha, for example #00FF00AA
- `Utils.lerp`'s table support (now handled by TableUtils.lerp) is more strict and will error if table sizes are mismatched

This commit needs SO MUCH TESTING before it's ready. So please, give it a shot. Thank you.


NyakoFox:
If nobody has any problems with this PR, it'll be merged soon. I'll give it a few more days.

Rudxain:
Not sure if this is relevant, but I was planning to add this (verbatim copy from my branch):
```lua
--- `if`-`else` expression.
--- Equivalent to `condition and a or b`, if all values are truthy.
--- `condition ? a : b` (AKA "ternary operator") in other langs.
---
--- This exists for readability and correctness purposes.
---
---@generic T
---@param condition boolean
---@param a T
---@param b T
---@return T
function Utils.if3(condition, a, b)
    if condition then return a
    else return b
    end
end
```
In my branch, I've already replaced 1 instance of `a and b or c`. It's a small patch, so I won't mind if I have to start over

Hyperfae:
For some reason, my dark fountain pillar recreation is twice as wide on this branch. I'm gonna have to investigate that.
image

Code: https://github.com/Hyperboid/fountain_make_cine_test

Hyperfae:
I have found that it's an issue with the Utils.clampMap proxy function. Updating to use MathUtils.rangeMap and MathUtils.clamp resolves the issue.
```diff
diff --git a/scripts/objects/FMPillar.lua b/scripts/objects/FMPillar.lua
index 09f149b..aaefdd9 100644
--- a/scripts/objects/FMPillar.lua
+++ b/scripts/objects/FMPillar.lua
@@ -47,11 +47,11 @@ end
 function FMPillar:draw()
     love.graphics.push()
     love.graphics.translate(80, 720)
-    local height = Utils.clampMap(self.anim_timer, 0.2, 0.3, 3, 320)
-    love.graphics.scale(Utils.clampMap(self.anim_timer, 0.1, 0.2, 2, 1),1)
-    love.graphics.scale(Utils.clampMap(self.anim_timer, 6, 6.1, 1, 0),1)
+    local height = MathUtils.rangeMap(MathUtils.clamp(self.anim_timer, 0.2, 0.3), 0.2, 0.3, 3, 320)
+    love.graphics.scale(MathUtils.rangeMap(MathUtils.clamp(self.anim_timer, 0.1, 0.2), 0.1, 0.2, 2, 1),1)
+    love.graphics.scale(MathUtils.rangeMap(MathUtils.clamp(self.anim_timer, 6.0, 6.1), 6.0, 6.1, 1, 0),1)
     self:drawCoolEllipse(height, 100)
-    local inner_luma = Utils.clampMap(self.anim_timer, 3,4,1,0)
+    local inner_luma = MathUtils.rangeMap(MathUtils.clamp(self.anim_timer, 3.0,4.0), 3.0,4.0,1,0)
     Draw.setColor(inner_luma, inner_luma, inner_luma)
     self:drawCoolEllipse(height, 96)
     if DEBUG_RENDER then
```

NyakoFox:
> I have found that it's an issue with the Utils.clampMap proxy function. Updating to use MathUtils.rangeMap and MathUtils.clamp resolves the issue.

I might've solved the issue, give it a try?

Hyperfae:
Nope, still happens. Reversing the order of the MathUtils calls seems to work, though.
```lua
local function Utils_clampMap(val, min_a, max_a, min_b, max_b, mode)
    if min_a > max_a then
        min_a, max_a = max_a, min_a
        min_b, max_b = max_b, min_b
    end
    if mode and mode ~= "linear" then
        local range = MathUtils.clamp(MathUtils.rangeMap(val, min_a, max_a, 0, 1), 0, 1)
        return Utils.ease(min_b, max_b, range, mode)
    end
    return MathUtils.rangeMap(MathUtils.clamp(val, min_a, max_a), min_a, max_a, min_b, max_b)
end
```
Concerning that the behavior is different, though.

NyakoFox:
Actually no, that makes perfect sense. I've pushed the fix!

NyakoFox:
> Not sure if this is relevant, but I was planning to add this (verbatim copy from my branch):

I'm not sure how I feel about this yet, but you could probably get better docs from:

```lua
---@generic T, K
---@param condition boolean
---@param a T
---@param b K
---@return T|K
```

NyakoFox:
> If nobody has any problems with this PR, it'll be merged soon. I'll give it a few more days.

Last call -- I'll merge this tomorrow if nothing comes up.

NyakoFox:
Looks like we're overdue -- I forgot yesterday.

Let justice be done, though the heavens fall.

PR #404, opened 2025-09-02
Narrow Dark Fountain

Agent517:
A dark fountain with the "narrow" property will look like the weird thin one in chapter 4's first sanctuary. The testmod fountain room now has both versions.

https://github.com/user-attachments/assets/156c3e6c-c15a-41de-80e3-504c7bed1eca


NyakoFox:
LGTM, thanks!

PR #408, opened 2025-09-08
File loading is now managed by coroutines. Add fallbacks if FFI doesn't work. Ensure project syntax uses Lua 5.1. Fixed an error when spamming "reload".

i-winxd:
Three changes in one!

While this change should not have any noticeable changes in terms of the API visible to modders, the change is sort of heavy since it rewrites the entire logic used to load files - though this only impacts methods that are already well-encapsulated.

## File loading changes

Loading new files used to be done on a separate physical thread. This has been changed to be managed by a coroutine that yields very frequently. **The performance impact is minimal** and the game does not freeze.

This ensures the engine can run in the event multithreading cannot work. Note that HTTPS is still run on a separate thread. Technically, it makes sense to make HTTPS into a coroutine, however, it's not needed. If multithreading can't work here, the module won't run.

## FFI doesn't work -> catch it

Modules requiring FFI, when it can't load FFI, no longer render the engine unplayable effectively by disabling it and wrapping the logic that loads FFI in a pcall

## Ensure project uses 5.1 compatible syntax

Removed all `goto` statements. Bitwise AND is now bundled rather than assumed to have been packaged with lua.

## Fixed an error when "Reload" button is spammed

The reload button and CTRL+R will not do anything if the game's state is loading or blank.

## Demo

This version can successfully compile to love.js and is accessible [HERE](https://i-winxd.github.io/kristal-wasm-compiler/).

*NOTE: debug > reload* crashes on the web port but works fine otherwise.

### DIFF
--- a/conf.lua
+++ b/conf.lua
@@ -15,7 +15,7 @@ function love.conf(t)
     t.identity = "kristal"
     -- TODO: hmm
     t.version = "11.0"
-    t.console = true
+    -- t.console = true
     t.window.title = "Kristal"
     t.window.icon = "icon.png"
     t.window.width = 640
--- a/conf.lua
+++ b/conf.lua
@@ -15,7 +15,7 @@ function love.conf(t)
     t.identity = "kristal"
     -- TODO: hmm
     t.version = "11.0"
-
+    t.console = true
     t.window.title = "Kristal"
     t.window.icon = "icon.png"
     t.window.width = 640
--- a/luadoc_meta/.buildDocs.lua
+++ b/luadoc_meta/.buildDocs.lua
@@ -15,8 +15,8 @@ export.makeDocObject['variable'] = function(source, obj, has_seen)
     if (obj.type == 'variable') then
         for i,pair in pairs(source:getSets(ws.rootUri)) do
             if(pair.type ~= 'setglobal') then
-                goto CONTINUE
-            end
+            -- continue
+            else
             --print(obj.name, i)
             obj.defines[i].value = '???'
             if(SIMPLE_TYPES[pair.value.type]) then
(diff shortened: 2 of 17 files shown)


NyakoFox:
A lot of this is very opiniated changes, especially the entire changes to the asset loader. I personally do not think moving to a less flexible system is a good goal, and this would work better as a fork (since this is for your WASM stuff). When people do pull requests, we normally want them to ask us about design first -- whether it's a good idea to put time into programming it or not. For this case, you can use this for your fork, but I don't think this should be accepted.

I agree that the loading thread is a little strange at the moment, but I do not think giving up the thread system is what we want to do here.

You do mention that for 5.1 compatibility, we shouldn't use `goto`. LÖVE uses LuaJIT by default (NOT 5.1) so `goto` is perfectly valid syntax, and the 5.1 compatibility would only apply for the WASM port. I'm sure a few `goto`s could be knocked out (normally you shouldn't use it at all, but multi-breaks are one of the widely accepted actual use-cases of them) but I don't think we should shy away from LuaJIT features since we're targeting default LÖVE. The `bit` library stuff also falls under this -- it's a library function in LuaJIT and it's fine to use.

What I DON'T like using in LuaJIT is stuff like `table.clear()` which is injection into the standard Lua library... but that's unrelated entirely.

Something I'd be willing to accept is the CTRL+R spamming fix, but there's definitely better ways to do it than checking `<current_state>.loaded`, relying on it being `nil` for any other state... I'd also be willing to accept the FFI thing, because I've been meaning to make that optional anyways.

Additionally, there really should not be multiple changes in one PR. It causes a lot more back-and-forth than needed, and the commits, despite there being four of them, also do not only contain one change per each. There's a lot of comments everywhere, naming issues (`data_thing`) and some of the code is a little messy.

Because of all of this, I don't think we can accept this PR. Please feel free to have a fork for your WASM changes (that's what this is, I assume) but I don't think a lot of this can be merged into Kristal. This isn't to discourage you (or anybody) from contributing either -- but a lot of care has to be put into contributing to a project this size. We were loose with PRs before, and it caused more problems than we would've liked.

PR #413, opened 2025-09-22
"TIRED" and "AWAKE" will no longer display if the enemy was defeated

FireRainV:
Will now only appear if the enemy has more than 0 health.
To fix an issue where it would display even if tired percentage is 0 or if the enemy was defeated in 1 hit before it appeared.

(re-PR because I forgot to make it a different branch.)


NyakoFox:
Shouldn't you be checking if they're defeated instead of checking their health? An enemy doesn't necessarily have to die when they have 0 health (ex. scripted stuff).

ghost:
this is how you check if they're defeated, even if you disable it.

NyakoFox:
Yes, but again what if someone doesn't use health whatsoever and want to disable the messages? This is inflexible and an assumption which may not always be true.

ghost:
well, do you have a better idea?

ghost:
actually, I have an idea, I will make it a return function

ghost:
but, @NyakoFox , what should be the default value?
true or health > 0?

NyakoFox:
the latter

ghost:
done. it can be merged now, I tested it and it works.

PR #415, opened 2025-09-23
Added enemy.show_hp

FireRainV:
Whether to show remaining health percentage or "???"

Just like in the fight against the Knight


NyakoFox:
I feel as if maybe this should be a callback rather than a hardcoded boolean... I'll push something in a bit to emulate this.

NyakoFox:
Done in d5a0b2e76b261debea392743dd1aea327af8714f.

PR #416, opened 2025-09-23
Implemented fixed Rude Buster damage calculation & new Chapter 3+ effect

HmmNoPls:
Rude Buster and Red Buster damage calculation for Chapter 3+ has been implemented.
Additionally, Rude Buster in Chapter 3+ has a slight change where the beam particles upon impact travel farther if the player lands a near-perfect or perfect hit. This has also been added.
Both of those can be toggled with the config option `oldRudeBuster`

### DIFF
--- a/configs/chapter1.json
+++ b/configs/chapter1.json
@@ -15,6 +15,7 @@
     "oldTensionBar": true,           // Syle of TP bar
     "oldUIPositions": true,          // Old positions of various UI elements
     "oldGameOver": true,             // The game over screen used in chapter 1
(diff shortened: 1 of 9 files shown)


NyakoFox:
Maybe this could be improved by removing all of the chapter 1/2 specific code for "pressed" & just breaking it the same way DR does, as the DR calculations were not new, just fixed.

It seems like we majorly simplified the code solely so that it'd be less confusing (as the bug pretty much just made it "either you press and get extra damage, or you don't") but it doesn't really have to be two different systems, probably just a single if somewhere.

Either way, for now this is good, thank you!

PR #417, opened 2025-09-23
Fixed a crash that would occur if a party member attacks without a weapon

ghost:


NyakoFox:
Shouldn't party members always have weapons? The entire engine assumes that, does it not?

ghost:
no, you can get rid of the weapon if the weapon doesn't exist or with a set weapon command.
I don't see why not fix it, it's a very simple fix.

PR #419, opened 2025-09-24
Fixed a bug where Recruit won't load a "false" boolean

FireRainV:
It will cause "LOST" to appear after every reload.


ghost:
this is an important bug fix, if you make an Enemy lost and you reload your save, they'll be recruit-able again.

PR #428, opened 2025-09-29
Fixed a mistake I made with the custom mercy display

FireRainV:
my bad.


ghost:
This is a critical fix.

PR #429, opened 2025-09-29
Changed "button.selectable" to "button.usable"

FireRainV:
I didn't know that button.selectable was used when the battle ends.
It caused the buttons to go gray when the battle ends.

I changed it to usable to fix that issue.


NyakoFox:
You need to test your PRs to make sure they work and don't cause problems. This is the second followup bugfix PR that you've made in the past hour. This is very common with your PRs.

Please test your commits before opening PRs. I don't want to see these as often. Ideally never.

ghost:
Sorry, I just didn't notice it at first.

ghost:
I kind of having the same issue when I'm updating MGR.
I only noticed this issue when I attempted to apply the update to MGR light battles.

PR #432, opened 2025-10-01
In the dark transition in the test mod, the child will be added to Game.stage

FireRainV:
The reason for this change is because when you load certain maps, it will display certain tilesets for a frame or two if the transition is added to the world instead of to the stage.

### DIFF
--- a/mods/_testmod/scripts/world/cutscenes/enterdark.lua
+++ b/mods/_testmod/scripts/world/cutscenes/enterdark.lua
@@ -27,7 +27,7 @@ return function(cutscene)
     end
     transition.layer = 99999
 
-    Game.world:addChild(transition)
+    Game.stage:addChild(transition)
 
     local waiting = true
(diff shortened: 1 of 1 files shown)


NyakoFox:
Why would this matter? It's the test mod.

ghost:
maybe if anyone wants to use that file for creating their own world dark transition, at least it should be without bugs.

NyakoFox:
nobody should ever reference the testmod, and while this is technically a bug in the testmod, i do not really want to encourage people to look in it anyways. they should NOT copy from that file. i have always said this and been strict about it, i even had to add a warning when you open the mod. nobody listens unfortunately. but if i accept this, i fear it'll look like me encouraging this. which i am not. so unfortunately i'm going to have to close this for now. maybe i'll sneak it in at a later date with a comment attributing it to you, but hopefully that gets across where im at.

PR #434, opened 2025-10-02
Tired and awake accuracy

AcousticJammYT:


ghost:
this seems way better than using health, nice @AcousticJammYT

AcousticJamm-dev:
It's also more accurate. The way your change worked, the text doesn't show up if the hit defeats. However, that is not the case in Deltarune.

PR #436, opened 2025-10-02
(Some) Accuracy for the right side shop panel 

MrOinky:
This PR changes the shopbox panel on the rightside to use the deltarune formula for expanding/shrinking (now toggled with the `Shop.expand_box` variable)
Additionally, heights and widths used in calculations for the shopbox and UIBox are floored before being passed into draw to negate the jittering effect that fractional parts could cause -- only height flooring should have been needed to fix this specific case, but i went through the widths as well, because it might as well be future proofed now incase a mysterious side-expanding panel appears or anyone mods that in themselves and encounters this (and, i've not observed any side effects from changing this when checking different UIBoxes in the engine from before and after this change, just in case it could have had unintended consequences)

### DIFF
--- a/src/engine/game/common/uibox.lua
+++ b/src/engine/game/common/uibox.lua
@@ -43,21 +43,24 @@ function UIBox:draw()
     local top_width   = self.top[1]:getWidth()
     local top_height  = self.top[1]:getHeight()
 
+    local width = math.floor(self.width)
+    local height = math.floor(self.height)
+
     local  r, g, b,a = self:getDrawColor()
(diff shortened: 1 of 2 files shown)


NyakoFox:
This mostly looks good, although I wonder if it's worth deduplicating a bunch of the `math.floor`s in UIBox?

MrOinky:
Ah you're right, i'll get on that in a minute

MrOinky:
hopefully this should be better now!

NyakoFox:
LGTM, thank you! This has annoyed me (and others) for a while so this is a great fix to have in the engine.

PR #437, opened 2025-10-03
Added EnemyBattler.getCheckText()

Simbel0:
This PR basically moves the code that sets up the Check text from `onAct` to its own function called `getCheckText()`

`getCheckText()` has two arguments:
- `battler`: the PartyMember that started the act, like onAct has.
- `separate`: Optional. If true, it will both return "* NAME - " and self.check separated instead of combining them together like it does usually.

May sound a bit overcomplicated? But this is basically what I suggested a while back on the server.

Also added an use case in the example mod as I believe learning there's a function that can override the check text is a good thing for new Kristal users.
There's also a use case in the _testmod for fun.

What about `onCheck()`? It's still here. _testmod now has a use case for it. You can easily see how it could be moved into `getCheckText()` but it doesn't feel right to me... So I guess it's still here for organization's sake?
If asked, I can just remove it altogether.

### DIFF
--- a/mods/_testmod/scripts/battle/enemies/virovirokun.lua
+++ b/mods/_testmod/scripts/battle/enemies/virovirokun.lua
@@ -12,6 +12,8 @@ function Virovirokun:init()
     self.asleep = false
     self.become_red = false
 
+    self.nb_checks = 0
+
     self:registerAct("Tell Story", "", {"ralsei"})
     self:registerAct("Red", "", {"susie"})
     self:registerAct("", "", nil, nil, nil, {"ui/battle/msg/dumbass"})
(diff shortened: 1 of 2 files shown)


NyakoFox:
I don't think "getters" should ever have side effects, so the separate onCheck is good to keep.

I feel like there shouldn't be any fancy "try to handle the name automatically" stuff... my idea of this function would pretty much just be doing the default behavior in EnemyBattler, and if something wants to return something else, then that's up to them:

```lua
function ThisEnemy:getCheckText(battler)
    return "* THIS ENEMY - No idea."
end
```

A lot of the time, trying to make a complex system that handles things for you often limits what people can actually do, and obscures what's actually happening -- `getCheckText` can be implemented as a simple override.

If you make these changes, please revert the example mod code -- I'm still unsure what we actually do with the example mod, but I generally don't want to accept PRs which do change it.

Simbel0:
Done, I think?

NyakoFox:
Looks good -- the testmod would technically need some fixes with the check text but it's the testmod so I'm not sure if I really care much. Thanks!

Issue #439, opened 2025-10-04
Chapter 4 mods are missing a 3rd page of item storage in save points


FireRainV:
In Deltarune, you get extra 24 slots for items in your storage in Chapter 4.

Hyperfae:
Wait, when was this fixed?

ghost:
https://github.com/KristalTeam/Kristal/pull/440

PR #440, opened 2025-10-04
Customizable Storage Pages

FireRainV:
You can now choose how many storage pages you'll have in the dark world.

In a Chapter 4 mod, you'll have 3 pages.

### DIFF
--- a/configs/chapter1.json
+++ b/configs/chapter1.json
@@ -1,5 +1,5 @@
 {    
-    "enableStorage": false,          // Extra 48-slot item storage
+    "storagePages": 0,               // The amount of extra storage pages
     "enableRecruits": false,         // Enable recruit messages and menu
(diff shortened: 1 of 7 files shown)


ghost:
fixes https://github.com/KristalTeam/Kristal/issues/439

SylviBlossom:
It would be preferable if the option configured the number of storage slots, rather than just the number of storage pages (registering it with the options of `0, 24, 36` to match base Deltarune values).

Also, it would be nice to smoothly deprecate the `enableStorage` option. Maybe an extra high-priority condition of:
```lua
    if Game:getConfig("enableStorage") ~= nil then
        Kristal.Console:warn("Using deprecated mod option 'enableStorage', switch to 'storageSlots' option instead")
        self.default_storage_slots = Game:getConfig("enableStorage") and 24 or 0
    end
```

Issue #433, opened 2025-10-02
There're a couple of bugs with the Dark Transition Animation


FireRainV:
1. There seems to be a leftover print command, because the console is printing numbers.
2. The white particle effects that appears while the party is changing forms is misaligned, especially on Kris.
3. There seems that there's no way of choosing the party x coordinate on the map when they finish their drop, only their y. (I think, correct me if I'm wrong.)

ghost:
Another bug, if you're only using 1 member, it will always be to the right instead of in the middle.

ghost:
fixed: https://github.com/KristalTeam/Kristal/pull/442

PR #450, opened 2025-10-09
You can now have any amount of characters in a dark transition

ghost:
In addition, if you have an odd amount of characters, the last in the crew will be at the middle. (Which is also accurate to Deltarune when Kris is on their own)

Here's a video for showcasing:
https://github.com/user-attachments/assets/d9b89ff6-38ee-475b-b8a6-77c48ba9b071


ghost:
fixes https://github.com/KristalTeam/Kristal/issues/433, most of it was already fixes by NyakoFox .

PR #451, opened 2025-10-11
Fixed get line intersection

TheSkerchh:
I needed it for a project so I fixed the function, this may be useful for projects such as the one I am working on in the future.
Stuff like shooting a line up and checking if it hit another line.

I used a formula I found [here](https://en.wikipedia.org/wiki/Line%E2%80%93line_intersection#Given_two_points_on_each_line)

Tested it with a few cases and all the results were correct.


SylviBlossom:
What cases does this fix? I haven't been able to find any inconsistencies between this and the original function where this solves an issue it had. I notice a "lines are the same" message has been added, but the original function also seems to handle that case - it just refers to them as "parallel".

This implementation also causes intersection with a 0-length line (start and end points equal) to return `nan`, but that could easily be fixed.

TheSkerchh:
Interesting, iirc I tested the original function and it was quite inaccurate, but there is a chance I got one of the arguments wrong in the testing.

TheSkerchh:
Alright so, this function sucks.

I've made a better version than this one which I'm gonna PR in a bit, gonna close this PR.

Issue #334, opened 2025-02-03
Memory leak issue with Kristal


FireRainV:
The more you unload or reload, even the Kristal main menu, when attempting to load a mod, the loading times and memory usage increases on a crazy level. The only temporary fix is to close and reopen Kristal.

ghost:
Are there any updates on this? This is such a critical issue with Kristal currently.

ghost:
@NyakoFox , @SylviBlossom  I actually found the source of the memory leak.
It's the function `Registry.initialize()` that being called at `Kristal.clearModState()`.

I tried commenting it and disabling crashes temporarily. And no leak! mods loaded instantly, no matter how many reloads I did.
I tried to fix it myself, but I couldn't manage, so I leave it to you guys as you know this stuff more than I do.

ghost:
I'm guessing that the Registry isn't being cleared properly before being initialized again.

ghost:
Fixed at https://github.com/KristalTeam/Kristal/pull/453

ghost:
Why did it got reopened? I already fixed the leak.
Is my fix not good?

AcousticJamm-dev:
Nyako mentioned this in the server so it's fair that you don't know, but she doesn't like closing issues for unmerged solutions.

Edit: My mistake. She commented it on one of your issues that you closed yourself.

Hyperfae:
It's reopened because the issue is still present on mainline Kristal (not your fork).

If a pull request fixes an issue, all you have to do is put "Fixes #[issue number]" in the PR description.

ghost:
That's it!
The most bothering issue with Kristal has been applied!
I was glad to be the one who helped fix it.

PR #453, opened 2025-10-13
Fixed Kristal's memory leak!!!!

ghost:
When clearing mod state, assets will now also get cleared (for some reason, they didn't), Game state will no longer be reset (I didn't find any issue with it not resetting, [Edit: it left leftover variables, but they'll now get cleared.]) and the registry won't be initialized again.

I check every scenario:
1. Adding new or removing an item or an asset like a sound effect, and then reloading, the item or sound got added or removed respectively.
2. If the game crashes due to a mod error, everything still works when you go back to the menu, fix the error, and everything works as expected.

@NyakoFox , I actually did it! I fixed the most annoying issue with Kristal!

### DIFF
--- a/src/engine/registry.lua
+++ b/src/engine/registry.lua
@@ -120,16 +120,9 @@ end
 ---@return boolean
 function Registry.restoreData()
     if self.saved_data then
-        local chapter = Kristal.getModOption("chapter") or 2
-        Game.chapter = chapter
-        
         for registry,path in pairs(self.paths) do
             self[registry] = self.saved_data[registry]
         end
-        
-        -- force garbage collection
-        collectgarbage("collect")
-        
         return true
     else
         return false
--- a/src/kristal.lua
+++ b/src/kristal.lua
@@ -1147,12 +1147,16 @@ function Kristal.clearModState()
     package.loaded["src.engine.game.game"] = nil
     Kristal.States["Game"] = require("src.engine.game.game")
     Game = Kristal.States["Game"]
+    Game.chapter = 2
 
     Kristal.setDesiredWindowTitleAndIcon()
 
     -- Restore assets and registry
     Assets.restoreData()
     Registry.restoreData()
+
+    -- force garbage collection
+    collectgarbage("collect")
 end
 
 --- Exits the current mod and returns to the Kristal menu.


ghost:
Okay, I think that's everything.
I've put it temporarily in darft due to "Game" variables not getting cleared, but now I fixed it and everything works.

NyakoFox:
Why does this remove reloading "Game" when you exit a mod?

ghost:
> Why does this remove reloading "Game" when you exit a mod?

What does it remove? Can you point to the code which does that?

ghost:
Oh, you mean that?
image

Because it will clear the registry while the game is loading, which will result in a crash.
So instead, I will just be clearing all the values of Game.

ghost:
Is there anything else you want to know?
You can test this PR yourself, I tried doing 1000 reloads (using a macro), and it was still clean, loaded mods instantly and the RAM usage was as usual.

ThePlasticPotato:
I would say that a proper profiler output might be better than just mentioning an arbitrary reload count- this change seems like it could break a lot of things for something without evidence to show why it solves the problem in the first place.

There is an existing lua profiler that Hyperboid used to diagnose this issue a while ago, so I'd recommend trying that.

ghost:
> I would say that a proper profiler output might be better than just mentioning an arbitrary reload count- this change seems like it could break a lot of things for something without evidence to show why it solves the problem in the first place.
> 
> There is an existing lua profiler that Hyperboid used to diagnose this issue a while ago, so I'd recommend trying that.

I used `print(collectgarbage("count"))` to check the RAM usage, and once I did this change, it didn't go up anymore, it stayed at stable amounts.
I checked every scenario, as I said at the top of the thread, nothing was broken.

What is that lua profiler Hyperboid did?

MrOinky:
Did some observing of the RAM with these changes and the changes to Game specifically do seem to keep the RAM usage stable (I won't pretend I know exactly Why this happens)
However! You shouldn't be messing with the registry here, the registry should ALWAYS be reinitialised when you clear the mod state, otherwise it is **not** clearing the mod state. If there was a problem with regsitry initialisation then this would only be deferring it to arise... a few milliseconds later when the registry is initialised again anyway in `preInitMod()` (every reload). 

Also, the change for `Kristal.clearAssets(true)` isn't doing any harm but it's entirely redundant as `Assets.restoreData()` is already responsible for clearing assets here - I think that line should have been left as it was

ghost:
> Did some observing of the RAM with these changes and the changes to Game specifically do seem to keep the RAM usage stable (I won't pretend I know exactly Why this happens) However! You shouldn't be messing with the registry here, the registry should ALWAYS be reinitialised when you clear the mod state, otherwise it is **not** clearing the mod state. If there was a problem with regsitry initialisation then this would only be deferring it to arise... a few milliseconds later when the registry is initialised again anyway in `preInitMod()` (every reload).
> 
> Also, the change for `Kristal.clearAssets(true)` isn't doing any harm but it's entirely redundant as `Assets.restoreData()` is already responsible for clearing assets here - I think that line should have been left as it was

The reason I added `Kristal.clearAssets(true)` is because if you delete an asset and then reload, it won't clear it from the memory.
Adding this and then using `Assets.restoreData()`, makes sure only currently existing assets will be loaded.

ghost:
> Did some observing of the RAM with these changes and the changes to Game specifically do seem to keep the RAM usage stable (I won't pretend I know exactly Why this happens) However! You shouldn't be messing with the registry here, the registry should ALWAYS be reinitialised when you clear the mod state, otherwise it is **not** clearing the mod state. If there was a problem with regsitry initialisation then this would only be deferring it to arise... a few milliseconds later when the registry is initialised again anyway in `preInitMod()` (every reload).
> 
> Also, the change for `Kristal.clearAssets(true)` isn't doing any harm but it's entirely redundant as `Assets.restoreData()` is already responsible for clearing assets here - I think that line should have been left as it was

Also, I check the registry, everything seemed to work fine without that initialization line, even adding new stuff or removing stuff would update it correctly. So I'm not seeing what is the issue here.

MrOinky:
Could you show what you mean by `Kristal.clearAssets(true)` being needed to make `Assets.restoreData()` only load "currently existing assets"? My understanding of the code is that `Kristal.clearAssets(true)` is a shorthand for `Assets.clear()` and `Kristal.Mods.clear()`, the latter of which was already called, and the former of which gets called first thing by `Assets.restoreData()` anyway - I wouldn't expect `Assets.clear()`ing a second time to clear anything new from memory as it's either already gone or being held by something other than Assets (and it'd be good to know if a double call is affecting something for some reason?).

As for the registry, I said it before and i'll say it again - this function is `Kristal.clearModState()`, it is responsible for **clearing the mod state**. `Registry.initialize()` is called within this function because it **clears everything that the mod has added to the registry**. By removing it, all these files will continue to exist into the menu where the mod state is not supposed to exist!
You would not notice anything wrong if you only tested Ctrl+R because as i mentioned before as well, `Registry.initialize()` gets called anyway by `preInitMod()` to load (or reload) the Mod's registry files. So in other words, removing that line has *no bearing* on what this PR is doing or anything about the mod reload process, but *fundamentally breaks* what the purpose of the function is when it really needs to do it, like returning to the menu (including after errors!). Even if it didn't cause any immediate issues, this is a problem on principle as code that shouldn't exist on the Kristal menu (files held in the mod's registry) now do, and they might expect part of the Mod/Game state to exist, or something might not expect them to exist anymore. Furthermore and much more importantly, when registry scripts trigger errors, the script that caused the error isn't getting cleared after leaving the error handler (as it relies on `Kristal.clearModState()` doing what it says it does!), this sometimes (albeit rarely) crashes the whole engine immediately (love2d error screen) if it's allowed to do whatever it did before again on the menu.

ghost:
@MrOinky I tested every scenario, including errors, nothing from the registry nor an asset that the mod made is saved upon reloading or exiting to the main menu.
I said all of it in my opening post of this thread.
Maybe check this PR yourself and compare it to the main Kristal branch and you'll see what I'm talking about.

ghost:
fixes https://github.com/KristalTeam/Kristal/issues/334

ghost:
@MrOinky before, without the line `Kristal.clearAssets(true)`, if you were to load your mod, remove an asset and then reload your mod, the asset will still exist. You can try it yourself. Try to add a sound, then reload your mod, then delete it and reload your mod again. Then if you try to play it, it will actually play it, even though you deleted it, no matter how many reloads you do.

ghost:
@MrOinky , actually, I tested it again.
The reason the assets didn't get cleared when I tested it before I added the `Kristal.clearAssets(true)` call is because I removed the line that completely resets the Game state, which lead to it still be loaded.
So adding that line fixed that issue.

If I kept the Game state reset, it would lead to a crash.

ghost:
nvm, you are right, there was no reason for it being there.
I think for some reason, it didn't clear when I tested it until I added that line.
weird, because now it actually gets cleared when I tested it without that line.

Okay, I restored it to what it was.

NyakoFox:
I looked at this a few days ago and put off reviewing it. I also pulled Sylvi in to look at it with me.

Thank you for trying to fix the memory leak, but this PR is unacceptable in its current state.

`elseif Kristal.Loader.waiting == 0 and Kristal.getState()["state"] then`

Rhetorical question: what does this even do? Will someone looking at this be able to figure it out? No. This is a weird bandaid, especially checking the loader's `waiting` variable in an obscured way.

Not reloading `Game` is unacceptable. `Game` is directly controlled by the modder, despite being an engine state. We can't *not* reload it. The system already sucks, and was a hack due to old design, and we do NOT want to make it worse. That's going in the wrong direction.

This is yet another PR that, instead of fixing root issues, does bandaids that "work". Other people have pointed out that your changes here either don't make sense, or don't need to exist. If a PR has this much discussion, your response should NOT be "I tested it and it works," you should be listening to the people who, for free, with no obligation, reviewed your PR.

I am sure that this does solve some problems with leaking memory. You would not open this PR if it didn't. However, you are not supposed to stop at "it works". You are supposed to go further and fix the real problems.

I cannot accept this in its current state. Which is a shame, because the memory leak problem is a massive one which affects everyone. But this is not a solution yet.

ghost:
`elseif Kristal.Loader.waiting == 0 and Kristal.getState()["state"] then`, this line prevent certain issues and crashes, like attempting to reload during the intro animation of Kristal or spamming reload that leads to you going to the main menu before it loads the mod again.

I hear you. I know how much you want to make the project the best it can be, but, have you tried this PR yourself? Doing all the tests that need to be done to know if there're any issues? Because I spend lots of time tinkering with everything which could cause an issue and after doing all these changes, I think I got everything. Still, I suggest that you do your tests yourself on that PR.

Hyperfae:
> have you tried this PR yourself?

Seems like you still don't understand. The issue is not whether it works, but rather that it doesn't make sense. It _needs_ to make sense.

> I am sure that this does solve some problems with leaking memory. You would not open this PR if it didn't. However, you are not supposed to stop at "it works". You are supposed to go further and fix the real problems.

NyakoFox:
> `elseif Kristal.Loader.waiting == 0 and Kristal.getState()["state"] then`, this line prevent certain issues and crashes, like attempting to reload during the intro animation of Kristal or spamming reload that leads to you going to the main menu before it loads the mod again.

I know exactly what it does, which is why I prefaced it with "rhetorical question". I was not asking for an explanation, I was saying that it's difficult to understand and obscured, explained right after the question.

ghost:
> > `elseif Kristal.Loader.waiting == 0 and Kristal.getState()["state"] then`, this line prevent certain issues and crashes, like attempting to reload during the intro animation of Kristal or spamming reload that leads to you going to the main menu before it loads the mod again.
> 
> I know exactly what it does, which is why I prefaced it with "rhetorical question". I was not asking for an explanation, I was saying that it's difficult to understand and obscured, explained right after the question.

I added these comments to better explain it:
image

Do you think it's okay? or is it too long?

ghost:
> > have you tried this PR yourself?
> 
> Seems like you still don't understand. The issue is not whether it works, but rather that it doesn't make sense. It _needs_ to make sense.
> 
> > I am sure that this does solve some problems with leaking memory. You would not open this PR if it didn't. However, you are not supposed to stop at "it works". You are supposed to go further and fix the real problems.

Well, idk, even if it gets closed without merges, I would still use it in my fork, this fix is way too useful for me to give up upon. I have no other idea on how to fix it using the methods you all want.
I checked the `Registry.initialize()` call, I could not find anything wrong with it, and yet, it still causes a memory leak when called during this function.

MrOinky:
The problem i've been raising with removing the `Registry.initialize()` call is not that there is something *immediately* broken upon removing it, rather that you've not given a logical explanation behind *why* you removed it (another piece of code *continues* to run `Registry.initialize()` when the mod reloads as I have already mentioned!) or addressed the implication that has for the state of the engine when returning to the menu. The main menu is a state where you'd *expect* nothing from the Mod to be loaded except for `mod.json` and `preview.lua`, and that call in the function NAMED `clearModState` is there **because** it CLEARS the registry from the mod state BACK to the pure Kristal state like when you first load the engine. Even if removing it doesn't break things right now, that line was originally there For A Reason and removing it **will** undoubtedly result in some kind of issue sooner or later.

ghost:
I'm already clearing every single variable that mods change to Game in here:
image

And as you said, because `Registry.initialize()` is called somewhere else, it applies any changes to the registry, if it's adding, removing or clearing it in the main menu or mod.

It doesn't seem to cause issues there.

MrOinky:
`Registry.initialize()` is called in `preInitMod()` on RELOAD (and for a separate purpose). That is not called when exiting to the menu, ergo **the registry is not reinitialized on the menu anymore** and registry items from the mod state flow over into the menu. `clearModState()` is expected to clean all that up, so should be **doing what it says it does**, which it doesn't without that line.

ghost:
@MrOinky and @NyakoFox , can you give this a look now? I did some changes.
The Game State will now get reset again, and I found a replacement for `Registry.initialize()` that won't cause a memory leak.

MrOinky:
This looks better on the Registry front - i'll suggest a few small tweaks about how this works though:
Since this is based on how `Assets.restoreData()` works (by the looks of it), i'd recommend following how Assets does the saving thing too (Assets.saveData()) and copy that more closely for the Registry - including storing the saved data in `Registry` itself - this should accomplish the same effect but with code between the two being much more consistent
Also, this doesn't clear out any potential extra keys in the registry, but neither did Registry.initialize() itself by the looks of it -- I only mention this because I know people have made custom registries before (the normal registry shouldn't end up creating new keys) and while i'm not sure how they implement them i'm guessing it involved new keys in the registry? The problem (if there is a problem) predates this PR by the looks of it and might be thing to fix separately but while i'm noticing it I might as well raise the issue

NyakoFox:
> > > `elseif Kristal.Loader.waiting == 0 and Kristal.getState()["state"] then`, this line prevent certain issues and crashes, like attempting to reload during the intro animation of Kristal or spamming reload that leads to you going to the main menu before it loads the mod again.
> > 
> > 
> > I know exactly what it does, which is why I prefaced it with "rhetorical question". I was not asking for an explanation, I was saying that it's difficult to understand and obscured, explained right after the question.
> 
> I added these comments to better explain it: image
> 
> Do you think it's okay? or is it too long?

The comments do not help, no. The problem is, still, that this does not explain things.

It explains what the result is, yes, but not WHY that code works.

Nonetheless, this back-and-forth has gone on long enough, so I've done it myself in 64e52c6a3fe3f941e2aa0edb01419418ec5d7396.

ghost:
@MrOinky , extra/custom keys are saved at the mod/library, not at the registry directly, which means once you exit the mod, it all gets cleared.
In addition, the way it works in Assets isn't the same, since you are going to only need to save the registry once when Kristal boots and restore it upon clearing the mod state. Assets have both a save and load functions, which don't fit to the way the registry functions.

@NyakoFox you haven't said anything about the new changes I did.

NyakoFox:
> @NyakoFox you haven't said anything about the new changes I did.

I trust MrOinky's judgement, I haven't needed to say anything.

MrOinky:
@FireRainV I'm not quite sure what you're seeing, but I can assure you if you look for all references to `Assets.saveData()`, it gets just a single call in loadstate.lua (the first "Kristal" loading screen, engine boot only), so it's situation is the same as Registry - saved once on engine load, and always restored back to that one state later - so they should ideally be handling that as similarly as possible
(I'll trust what you're saying on the custom registry front, as that does sound more reasonable than what I was thinking anyway)

ghost:
@MrOinky check it out now, I did it like you wanted.

MrOinky:
Yep, that looks better!
One last thing for cleanliness: Move setting `Game.chapter` to `clearModState` and just set it as `Game.chapter = 1`, no need to check the mod's chapter value as it's not going to exist

ghost:
@MrOinky here, I did the changes you requested.

MrOinky:
Checked this yesterday but forgot to actually say everything looks good 👍

PR #454, opened 2025-10-13
Enemies will now always show TIRED message by default

ghost:
It caused an issue where if you disable tired by damage, it wouldn't show it even if you make the enemy tired by an ACT.

The Modder can set it to "false" themselves if they need to.

### DIFF
--- a/src/engine/game/battle/enemybattler.lua
+++ b/src/engine/game/battle/enemybattler.lua
@@ -612,7 +612,7 @@ function EnemyBattler:getEncounterText()
         return self.spareable_text
     end
 
-    if type(self.low_health_percentage) == "number" and self.low_health_text and self.health <= (self.max_health * self.low_health_percentage) then
+    if type(self.low_health_percentage or false) == "number" and self.low_health_text and self.health <= (self.max_health * self.low_health_percentage) then
         return self.low_health_text
 
     elseif self.tired_text and self.tired then
@@ -852,7 +852,7 @@ function EnemyBattler:onHurt(damage, battler)
     end
     self:getActiveSprite():shake(9, 0, 0.5, 2/30)
 
-    if type(self.tired_percentage) == "number" and self.health <= (self.max_health * self.tired_percentage) then
+    if type(self.tired_percentage or false) == "number" and self.health <= (self.max_health * self.tired_percentage) then
         self:setTired(true)
     end
 end


ghost:
an easier way actually, if you want to prevent it from showing on tired by damage, simply set the percentage to a negative number like -1

Hyperfae:
> an easier way actually, if you want to prevent it from showing on tired by damage, simply set the percentage to a negative number like -1

...what? Could you explain how that works? And, why not just override shouldDisplayTiredMessage to return false?

ghost:
> > an easier way actually, if you want to prevent it from showing on tired by damage, simply set the percentage to a negative number like -1
> 
> ...what? Could you explain how that works? And, why not just override shouldDisplayTiredMessage to return false?

I'm talking about doing this: `self.tired_percentage = -1`.
The reason I don't want to override it, is because if you were to set it to 0, when performing an act which makes the enemy tired, the TIRED message won't show up.

Hyperfae:
That would prevent tired by damage in general, not just the message. Same applies to setting it to 0.

ghost:
not true.
if you were to set it to 0, if the enemy dies, it will say TIRED right before running away.

Hyperfae:
And you've verified that on the latest Kristal commit?

AcousticJamm-dev:
Even if you were to set the tiredness percentage to a negative number, wouldn't you still be able to get there with enough overkill damage?

ghost:
> And you've verified that on the latest Kristal commit?

yes

ghost:
> Even if you were to set the tiredness percentage to a negative number, wouldn't you still be able to get there with enough overkill damage?

you can't deal overkill damage, the enemy health is maxed at 0.

ghost:
> Even if you were to set the tiredness percentage to a negative number, wouldn't you still be able to get there with enough overkill damage?

nvm, you're right.
then set it to -math.huge instead.

ghost:
no, actually! I have an even better idea!
basically, `make it self.tired_percentage >= 0`, that way, `-1` will work.

Hyperfae:
0 _should_ work, though. #412 was supposed to do that.

ghost:
lets just use `-math.huge` and be done with it, no way it can break that way.

Hyperfae:
But that's unintuitive, which, in case it isn't obvious, is a huge problem.

ghost:
do you have better ideas?

Hyperfae:
Re-doing #413 to actually do what it was supposed to do.

ghost:
let me check something real quick.

ghost:
okay, after checking, it seems that even if you 1-shot an enemy before the TIRED message appears, it will still display TIRED upon them running away.

ghost:
so doing health > 0 won't do it

ghost:
> lets just use `-math.huge` and be done with it, no way it can break that way.

just do that, I can't think of a better option without adding a useless boolean.

ghost:
actually, I have an idea

NyakoFox:
> so doing health > 0 won't do it

This is the exact reason why we removed this check. Because it was inaccurate to DR.

> > lets just use `-math.huge` and be done with it, no way it can break that way.
> 
> just do that, I can't think of a better option without adding a useless boolean.

I'm not sure if you realize this, but a massive part of Kristal is the **API design**. I'm getting really annoyed by the "if it works, it works" mentality. No, we are NOT using `-math.huge`.

Spend time thinking about what would be best for the user. Things should be clear and easy to understand. A note somewhere like "for this specific behavior, use this `-math.huge` workaround" is NOT what we want to do. This is a fangame engine, usability is extremely important.

Stop using discussions as a chat room. Stop posting multiple times in a row like this.

ghost:
Okay, that should do it, if you set self.tired_percentage to nil, the enemy will no longer be tired by damage, and it doesn't have to do with math.huge this time.

ghost:
well, this change is now pointless due to https://github.com/KristalTeam/Kristal/pull/457

Issue #423, opened 2025-09-26
There's no option to choose the load order of libraries in the mod


ghost:
it's hardcoded, sometimes feels random after every reload.

MrOinky:
The load order of all files is effectively random (because `love.filesystem.getDirectoryItems()` gives stuff in a non-deterministic order, I believe?) BUT libraries do already have the option of specifying optional dependencies in `lib.json`, which is specifically for this case, making a library load after the other specified libraries

ghost:
okay, but even with optional dependencies, you need to tell the creators to put them there.
There should still be an option in `mod.json` to order the libraries yourself.

ghost:
done in https://github.com/KristalTeam/Kristal/pull/455

NyakoFox:
stop closing issues for unmerged PRs

NyakoFox:
This is not planned. If a load order is needed in this way, then you're doing something wrong. I assume you implemented this for the `engine-fixes` library, which clobbers entire classes. Your hook files will only cause problems, and I urge you to fix the library instead of requesting a feature to "fix" your problem.

ghost:
> This is not planned. If a load order is needed in this way, then you're doing something wrong. I assume you implemented this for the `engine-fixes` library, which clobbers entire classes. Your hook files will only cause problems, and I urge you to fix the library instead of requesting a feature to "fix" your problem.

Actually, all the fixes that I had in 'engine-fixes' library were already applied to base Kristal.
That library has been deleted and MGR no longer requires it.

The only thing that I moved from 'engine-fixes' is this: https://github.com/KristalTeam/Kristal/pull/471 .

PR #455, opened 2025-10-14
Customizable libraries order

ghost:
There's a new option in mod.json, which lets you give priority to libraries load order.
By default, if a library with priority has dependencies, it will load them first before it loads itself.
If a library isn't on the list, it will load them afterwards, like usual.

### DIFF
--- a/src/engine/game/game.lua
+++ b/src/engine/game/game.lua
@@ -510,15 +510,6 @@ function Game:load(data, index, fade)
         end
     end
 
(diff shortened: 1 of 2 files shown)


ghost:
fixes https://github.com/KristalTeam/Kristal/issues/423

ghost:
@MrOinky what do you think of this method I did? would you change anything? should I keep the changes I made for mod.json for the mod template?

PR #456, opened 2025-10-15
The candy box can no longer kill

ghost:
@NyakoFox , maybe reconsider this change? I don't think Toby would make you die from eating the candies if you somehow were to be on 1 HP.
In Undertale, if you were to eat a bad memory on very low HP, it would restore your HP instead, so I think it would fit the same for the candy box.

Like, I don't think it's fair for the player that they'll have an unexpected death because they used a "safe" item and then lose progress because of that.

### DIFF
--- a/data/items/light/box_of_heart_candy.lua
+++ b/data/items/light/box_of_heart_candy.lua
@@ -26,10 +26,7 @@ end
 
 function item:onWorldUse()
     Game.world:startCutscene(function(cutscene)
-        if Game.party[1].lw_health <= 1 then
-            Game:gameOver(Game.world.player.x, Game.world.player.y)
-            return true
-        else
+        if Game.party[1].lw_health > 1 then
             Game.party[1].lw_health = Game.party[1].lw_health - 1
         end
         cutscene:text("* (You unhesitatingly devoured\nthe box of heart shaped\ncandies.)")


NyakoFox:
I already told you my thoughts. The item that deals one damage should deal one damage. Don't falsely equate it to the bad memory, which is a special, magic item. This item is a box of chocolate that deals one damage to the player.

We are ALREADY making an uncomfortable design decision making it not just set your HP to 19 (which is what it does in DR). I do not want to make it worse.

ghost:
but, should it cause a game over?

NyakoFox:
Dealing one damage should cause a game over. If you want different behavior for your mods, then sure, replace it in your mod. Otherwise, this doesn't matter.

Keep in mind that the light world is not Undertale. The light world in DR is not meant for complex gameplay. It's a one-off joke item that we reimplemented. If you want something else, add something else yourself.

NyakoFox:
I forgot to say this, but it's important so I'll say it in a follow-up comment instead.

To clarify, in this engine, we want to make little to no decisions FOR Toby Fox. People do not use this engine for "Ally's version of DELTARUNE", they use it because it's an accurate DELTARUNE engine. If we start making decisions that aren't in DELTARUNE, then it's no longer DELTARUNE. It's *our* version of DELTARUNE. And that, to me, is unacceptible.

That's why "1 damage" is already uncomfortable to me, but at least there's a good reason WHY we did that.

That's why I'm very uncomfortable with the inclusion of the "light world game over" that I did years ago, because we made some random decision because I thought it was cool.

This engine is supposed to be accurate to DELTARUNE. I have no place making decisions when this is Toby Fox's game. It would be insincere to call this an accurate engine if I started making decisions like that.

ghost:
Okay.
I just thought it would make more sense that the Candy Box will stop hurting you if you already have 1 health to prevent a game over. Seems like something Toby Fox would do if it you somehow could get your health to 1.

Why feel uncomfortable about that? It'll be impossible to get 100% accuracy, or even if we could, it's not necessarily the best option. Yes, we try to mimic deltarune, but in cases like this, where in DR it sets your HP to 19, because there's no other way of taking damage in the chapter so it's not a big deal, why not make it the most reasonable decision?

ghost:
I leave it up to you, this is such an insignificant change that I wouldn't really care if it won't get applied, I just thought it would be a better design decision. I already disabled that death in MGR as it would cause a mess with multiple party members.

PR #457, opened 2025-10-15
Rework TIRED message controls (with `setTired()`)

MrOinky:
The `setTired(bool, hide_message)` function now has a `hide_message` argument that can be used to hide TIRED (and awake!) messages, instead of needing to use global condition functions on the enemy (that were also unaware of the tired-setting context).
When getting tired from damage, the message will now also be automatically hidden this way if  `tired_percentage` is 0 (or below, for some reason), as it's fair to assume that mod authors expect `tired_percentage = 0` wouldn't result in a TIRED message on kill.
This also removes the tired/awake message condition functions as they are naturally superseded by this system

### DIFF
--- a/mods/_testmod/scripts/battle/enemies/virovirokun.lua
+++ b/mods/_testmod/scripts/battle/enemies/virovirokun.lua
@@ -23,8 +23,16 @@ function Virovirokun:init()
         description = "Red\nDamage"
     end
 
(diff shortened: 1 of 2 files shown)


NyakoFox:
Reviewing this made me tired. Ha ha . do you get the joke. i was lying it was just a joke

PR #470, opened 2025-10-28
Pacify with no visual effects config option (from Chapter 1)

ghost:
In addition, fixed the SnowGrave spell random calculation.


ghost:
I applied the change

PR #472, opened 2025-10-28
Adds Input Lag for accuracy

PKMNTrainerZion:
Adds the Deltarune-accurate 3 frame lag after pressing confirm in the attacking state.

### DIFF
--- a/src/engine/game/battle.lua
+++ b/src/engine/game/battle.lua
@@ -3389,7 +3389,7 @@ end
 ---@param key string
 function Battle:handleAttackingInput(key)
     if Input.isConfirm(key) then
-        Game.battle.timer:after(1/10, function()
+        Game.battle.timer:after(1/20, function()
         if not self.attack_done and not self.cancel_attack and #self.battle_ui.attack_boxes > 0 then
             local closest
             local closest_attacks = {}


ghost:
I just tested this change, and it completely breaks the ability to hit the target.
I think its better off without this.

NyakoFox:
Do not speak for us. You do not have authority here.

ghost:
?
I just said my opinion on this change, why being so hostile?

NyakoFox:
I am not being hostile, I am being strict. There have been times where someone weighed in to a PR and it got completely changed to what they wanted without my input.

ghost:
Just say that you want me to not say anything

PKMNTrainerZion:
> I just tested this change, and it completely breaks the ability to hit the target.

Could you be a bit more specific? I've tested this on multiple framerates while on stream with a couple of friends and they agree they don't see any issues. Is there something I'm not noticing? Btw, if you haven't played DR in a while, this may feel odd because Kristal doesn't have this by default. This is accurate to DR. That's why I'm asking.

ghost:
@PKMNTrainerZion the delay felt huge. I couldn't get any crits at all with this change, the delay is really noticeable.
I checked in Deltarune again, and it felt completely different than this change.
I'm not sure why, but maybe it was already accurate before this change? Maybe something with Kristal's battles was already correcting it?

ghost:
All I'm trying to say is that I think it's already accurate in its current state, before this change.
In Deltarune, the bar was active the moment I pressed Z, I didn't see any delay, same with Kristal.
With this PR change, pressing Z will make the bolt move forward more and then hit, which makes it too difficult to hit. You can have the bolt on top of the target and even if you press Z, the attack will just miss compared to before.

PKMNTrainerZion:
@FireRainV So I looked back at my coding, and I realized that you _are_ partly correct. Kristal doesn't have any delay, but Deltarune has a very, very small one. It's very hard to notice since it feels very fluid. I've changed the coding from 3 frames to only 2 frames of delay, as that was my mistake. I'm gonna fix the PR now.

ghost:
I think I can confirm that there's no input lag.
I just used slowdown in Deltarune (by setting the game to 3 FPS), and the attack hit the moment I pressed Z, there was no lag.

PKMNTrainerZion:
Look. I don't want to argue, but several people have said that there is a delay. I only wish to make the engine more accurate. Ultimately, I will leave this up to the developers.

ghost:
> Look. I don't want to argue, but several people have said that there is a delay. I only wish to make the engine more accurate. Ultimately, I will leave this up to the developers.

agree, better let the devs handle it.
We all want Kristal to be in the best state it can be.

NyakoFox:
> Deltarune has a very, very small one. It's very hard to notice since it feels very fluid. I've changed the coding from 3 frames to only 2 frames of delay

Where are you getting 2 (or 3) frames from? I'm fine with the delay, but I do want to make sure it's accurate.

PKMNTrainerZion:
To be honest, I'm unsure if it's actually implemented in the coding, or if it's a side result of something, but as of now, I'm kind of eyeballing it. I'll make a post on Game banana so I can get some help finding where this may be implemented in DR, but like I said, I'm not getting this with actual coding, so accuracy may be a bit faulty.

PKMNTrainerZion:
@NyakoFox So I found it in DR's coding, and funnily enough, I was correct. In _gml_Object_obj_dmgwriter_Create_0_ there a line that defines _delay_ as 2.

NyakoFox:
That's the damage number, nothing to do with attacking...

NyakoFox:
I do wonder if there's a better way to do this; making the bar location always jump to the "next" 30FPS target, which would introduce the 1-frame delay DR has, while also making it not jump "backwards" at higher targets...

ghost:
it will be very annoying when doing a slowdown though, trying to test the target, only for it to teleport.
It will cause too many unfair misses.

ghost:
it's like introducing the bug where for a few frames, it still uses the last loaded portrait of the character in a new text box before updating to the correct one.

ghost:
I'm really against this change.

AcousticJamm-dev:
@FireRainV I get what you mean, and it's possible there are a few issues with this change, but this is a change made for Deltarune accuracy, which Ally wants in this engine.

NyakoFox:
Once again, the point of this engine is to be DELTARUNE-accurate. NOT "Our improved version of DELTARUNE"-accurate.

It should jump forwards, yes. That's the nature of DT.

The change is unnoticeable as well. I don't think a 1 frame delay is the right implementation, but this is somewhat high priority to me.

NyakoFox:
As this is implemented in a different way than what we'd like, I will close this PR for now. Feel free to open an issue for attacking input delay, however.

PR #477, opened 2025-11-02
Fixed Party Member level-up inaccuracies

nitrotrixium:
Previously, the code increased the global level count once per party member, causing uneven level ups. I fixed that by putting the level up increment outside of the for loop.

### DIFF
--- a/src/engine/game/battle.lua
+++ b/src/engine/game/battle.lua
@@ -629,8 +629,8 @@ function Battle:onStateChange(old,new)
                 end
             end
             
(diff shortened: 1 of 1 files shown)


NyakoFox:
LGTM, thanks! You didn't add your username to the engine credits; I can commit that myself, what name would you like?

nitrotrixium:
> LGTM, thanks! You didn't add your username to the engine credits; I can commit that myself, what name would you like?

I'm abbe_abbington in the discord :>

PR #478, opened 2025-11-04
Old Dual Heal formula config

NellMonell:
In chapter 1, Dual Heal is calculated with "magic * 4", instead of "magic * 5.5" like it's done in chapters 2 and 3 (even if in chapter 3 it's unused)


NyakoFox:
LGTM, thanks!

PR #482, opened 2025-11-10
Check Description config

NellMonell:
Starting from chapter 3, practically every "Check" act had description "Useless analysis", aside from a few exceptions. This config adds that. Also I added that and my previously added config "oldDualHealFormula" to the mod creation config.


NyakoFox:
LGTM, thanks!

PR #483, opened 2025-11-11
Events now use parallax of parent layer

HmmNoPls:
Originally, events are unaffected by the layer's parallax settings


NyakoFox:
LGTM, thanks

PR #485, opened 2025-11-12
FiberScarf attack bonus accuracy

NellMonell:
Weirdly, specifically in chapter 3 and there only, the attack bonus of FiberScarf is set to 3 instead of usual 2.


NyakoFox:
they probably planned to change it, they changed it in chapter 3 and forgot to do it for 4. both chapters were developed at the same time so any changes like that had to be done twice, im surprised theres not more issues like this

PR #486, opened 2025-11-14
Select Multiple Waves debug menu

MrOinky:
(Completely repurposed PR because I'm Evil)

Adds a multiple wave selector to the debug menu (does what it says on the tin)
Also adds two cool new debug menu features to make that possible:
- Colored debug options (new last argument to `registerOption`, can be a function (dynamic color) or table (static color)
- Option to make debug options "fail" or otherwise play `ui_cant_select` instead of `ui_select` by returning `false` in their functions
Debug menus that have "selected" style elements will use the color feature as well (Spells, Party members & Music test)

### DIFF
--- a/src/engine/game/debugsystem.lua
+++ b/src/engine/game/debugsystem.lua
@@ -798,6 +798,9 @@ function DebugSystem:registerSubMenus()
         "[Start Waves]",
         "Start the selected waves.",
         function ()
+            if #self.selected_waves > #Game.battle:getActiveEnemies() then
+                return false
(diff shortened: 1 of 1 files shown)


MrOinky:
Since the old PR here was silly and Bad (an issue with one of my mods and not the engine 💥 ) I have completely swapped this out with something new (tangentically related to the old one)... this is totally how you use PRs haha... ha...

NyakoFox:
LGTM, thanks!

PR #488, opened 2025-11-21
Fixed an issue with pull request #483 (Event parallax)

HmmNoPls:
Map:loadObjects now multiplies the event's original parallax instead of overriding it with the layer's, fixing the issue of parallax set in :init of the event being cancelled


NyakoFox:
Thanks!

PR #490, opened 2025-11-28
Use Game:getFlag instead of directly indexing the flags table

Dobby233Liu:
In case someone hooks Game:getFlag for some funny reason (idk what)


NyakoFox:
LGTM, no idea why we were doing this

PR #491, opened 2025-12-01
Added Shop:isWorldHidden() method

DIAMONDDELTAHEDRON:
Game:update() now uses this to allow shops to control whether or not to hide the world. Useful for Ch3+4's overworld shops.

You can also disable the fade-in when exiting a shop by setting the leave option "fade" to false.

### DIFF
--- a/src/engine/game/shop.lua
+++ b/src/engine/game/shop.lua
@@ -571,7 +571,7 @@ function Shop:leaveImmediate()
 end
 
 function Shop:shouldFade()
-    return self.leave_options["fade"] or self:isWorldHidden() == true
+    return self.leave_options["fade"] or self:isWorldHidden()
 end
 
 --- *(Override)* Called whenever the player enters the TALK submenu.


NyakoFox:
I do wonder if we can implicitly check whether it should fade or not by default by using the "is world hidden" function, or maybe an override for like, should fade... the "leave options" thing really sucks especially when you're supposed to inherit Shop anyway... wtf was I doing.

DIAMONDDELTAHEDRON:
Fair enough. It probably should do that.

src/engine/game/shop.lua
```diff
@@ -570,6 +570,10 @@ function Shop:leaveImmediate()
     end
 end
 
+function Shop:shouldFade()
+    return self.leave_options["fade"] or self:isWorldHidden() == true
```
NyakoFox: What's the point of the `== true` here?

NyakoFox:
Thanks! (Feel free to open another PR to add your name to the credits...)

PR #493, opened 2025-12-03
SmallFaceText text now checks for actor:getFont()

HUECYCLES:
You can now use an actor's font (if any) in textbox reactions.


NyakoFox:
Looks good I think, thanks!

PR #494, opened 2025-12-07
Persist DamageNumber and RecruitMessage into the overworld

Hyperboid:
Since Deltarune does not separate battles from the overworld through object parenting like Kristal does, they don't need to do anything special to achieve this effect.


NyakoFox:
Kind of scary behavior here, but I guess this might be the best way... You should probably check for Game.world actually existing, though.

Hyperfae:
Alright, your suggestion has been added in commit 9376b0121b88db8320cf00767f67c0ee8adf3f98. This should be ready to go now.

NyakoFox:
LGTM, thanks!

PR #495, opened 2025-12-10
Item accuracies

NellMonell:
Some items from chapter 3 change their price in chapter 4. Also, "chapter 1" reactions to Pink and White Ribbons are actually used in chapters 3 and 4, which means reactions change only in chapter 2 for those items.


NyakoFox:
LGTM, thank you!

PR #500, opened 2025-12-27
Evict Inverse Vampirism (Make MirrorAreas respect object visibility)

Agent517:
Self-explanatory. A mirror shouldn't reflect something that's invisible, right?


NyakoFox:
It's a funny name but please name your PR's properly, and more importantly, commit messages.

PR #501, opened 2025-12-27
Fixed get line intersect(But actually this time)

TheSkerchh:
So I went back and tested it, the segment part of Utils.getLineIntersect seemed to mess up the results

image

All of these lines are supposed to intersect, but one of them is segmented which messes stuff up.

This updated function should fix it.

### DIFF
--- a/src/utils/utils.lua
+++ b/src/utils/utils.lua
@@ -1003,7 +1003,6 @@ function Utils.getLineIntersect(x1, y1, x2, y2, x3, y3, x4, y4, seg1, seg2)
             return false, "The lines don't intersect." 
         end
     end
-
     return x, y
 end
 


TheSkerchh:
I've tested it and it seems to work now(I'm using it for my own project and it seems to work every time)

TheSkerchh:
In the defence of this PR
(This will be using my purple soul library, which uses getLineIntersect to check for strings above and below the soul)


https://github.com/user-attachments/assets/75f9a48e-42ba-4acb-8954-f8b183251b20



In this video, Utils.getLineIntersect is used. You can see it is missing some inputs, as well as skipping strings from time to time, I believe this is an issue with the segmented line checks in the getLineInterset function flagging a false negative sometimes(This was definitely way more noticeable and annoying in an attack from the new wilter's wonderland update that I sadly can't show I think).

https://github.com/user-attachments/assets/f5b89f0c-4c2e-4b1b-ba79-652e55c51f41


In this video, I've changed the function to my own getLineIntersect(what's in my PR). You can see it is no longer missing inputs and making false negatives. This may be because I've changed the equation to find the intersection, or much more likely the changes I made to the segmented line checks, they should be much more reliable in the fixed function.

NyakoFox:
This is the third time we were shown this video; we've seen it. Please be patient when it comes to PR review; personal factors keep me from reviewing consistently. Either way, this can be merged now that we've had more time to look at it, but it needs its conflicts resolved.

PR #504, opened 2026-01-03
Restored debug tension to 200 in battles

ghost:
It makes it easier to cast and test SnowGrave.
200 is still being used in `function World:onKeyPressed(key)`

I'm not sure which of one you want, 100 or 200.


NyakoFox:
I changed it to 100 for a reason. If you need something else, do it yourself.

NyakoFox:
> I changed it to 100 for a reason. If you need something else, do it yourself.

I re-read this and realized my tone was off; I mean that if the modder needs a different value, they'd be better off setting it in the console.

PR #507, opened 2026-01-10
Fixed a draw glitch with the background of the Tension Bar

ghost:
Fixes an issue where the background of the tension bar would get cut off when getting TP.
(P.S., it's very noticeable if you play around with the foreground of the Tension Bar, and the fix is very simple.)


Hyperfae:
So here's a before and after. I've set `PALETTE["tension_fill"] = {1,0,1,0.5}` (semitransparent magenta) so you can see the difference.
image&nbsp;image
As far as I can tell, this pull request causes transparent fill colors to not function properly.

ghost:
@Hyperboid well, I didn't really think of someone using transparency. In that case, it's better to just leave it be, as the glitch isn't visible under normal circumstances anyways.
I will close this PR.

PR #508, opened 2026-01-11
Fix for Encounter:getSoulSpawnLocation()

ghost:
Fixes a bug where the main_chara would always pick their default actor instead of picking their current active actor.


NyakoFox:
LGTM, that's a tricky one.

PR #509, opened 2026-01-12
Fixed a crash that would occur when calling a class

ghost:
Fixes this:
`src/utils/classutils.lua:20: bad argument #1 to 'ipairs' (table expected, got nil)`

### DIFF
--- a/src/utils/classutils.lua
+++ b/src/utils/classutils.lua
@@ -17,7 +17,7 @@ function ClassUtils.getClassName(class, parent_check)
         end
     end
     -- If the class doesn't have a global variable, find the name of the highest class it extends.
-    for i, v in ipairs(class.__includes) do
+    for i, v in ipairs(class.__includes or {}) do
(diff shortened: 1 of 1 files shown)


ghost:
It can sometimes happen when you have table of tables in a class, and then attempting to call the class.

ghost:
example, if you try and do something like `print(Mod.libs["magical-glass"])` in the console (when using my lib), it will give you that error due to the existences of nested tables variables in the class. This change fixes this error.

Hyperfae:
The issue isn't even with the class system itself. It's with this function's handling of tables that have metatables, but _aren't_ using Kristal's class system. `ClassUtils.getClassName` assumes that the table is a Kristal class, while `isClass` doesn't rigorously check such.

ghost:
@Hyperboid you think you can find a better fix? I just managed to prevent the crash, probably there's a better way to handle it.

ghost:
I will close this since there's probably a better way of handing the error

PR #510, opened 2026-01-13
Disable box's corner flash with Simplified VFX

NellMonell:
"Simplify VFX" barely does anything in Deltarune. But what do you know, one of those "anythings" is disabling the little flash animation of ui boxes in dark style! This commit recreates that behavior.


NyakoFox:
Looks good to me!

At some point, we have to let UIBox specify the animation speed (as its extremely inconsistent in DR and actually relies on uh, how many boxes currently exist... seriously, go look at shops in DR.) but I haven't thought of a good way to do that yet...

PR #511, opened 2026-01-16
Better error upon incorrect tile layer format

Hyperboid:
Before, the error would only happen upon trying to draw the tiles. Since it's a string instead of a table, error happens.

Perhaps in the future we could support compressed formats.


NyakoFox:
LGTM, thanks! Explicit errors are always really good to have!

PR #513, opened 2026-01-18
Support new Tiled list properties

Hyperboid:
As of mapeditor/tiled#4002, properties can now be arrays. This pull request allows TiledUtils.parsePropertyList and TiledUtils.parsePropertyMultiList to support this format.


NyakoFox:
Looks good to me! And it took them long enough. We've wanted these for years.

PR #516, opened 2026-01-20
Fixes Border aspect ratio

PKMNTrainerZion:
Fullscreening with certain monitors causes a weird offset or graphical glitch with borders. This aims to fix that issue by manually aligning the border.


Hyperfae:
This creates a new `kristal.lua` in the root of the project. You should instead modify the existing one.

NyakoFox:
Closing awaiting proper PR (can't review as there's no diff)

PR #519, opened 2026-01-22
Spell and Item target getters

NellMonell:
Adds the "getTarget()" function to both spells and items, and replaces all checks of ".target" with the said function


NyakoFox:
Looks good to me -- I do see that Item is lacking doc comments but the rest of the file is so we can do that in one go some time.

PR #521, opened 2026-01-31
Dubious spell menu-related crash fixes

Dobby233Liu:
Haven't checked DR behavior


NyakoFox:
Needs conflicts fixed.

Also, does this... work? Like, the `tp_offset` stuff was moved inside of a conditional, which is a local variable, reducing the scope... and then something later, *out of scope* uses it?

NyakoFox:
LGTM, thanks!

PR #523, opened 2026-02-01
Use "local function" syntax for declaring Registry.iterScripts subroutines

Dobby233Liu:


Dobby233Liu:
This allows the functions to be named in tracebacks

PR #524, opened 2026-02-01
Slightly reorganize Character:jumpTo variables

Dobby233Liu:
Mainly just to prevent it from clobbering self.init


NyakoFox:
this function makes me upset whenever I remember it exists

PR #525, opened 2026-02-01
Add a lerp angle function to Math Utils

MaybeSamo:
It's pretty easy to add yourself, but it would be nice to have this as a built in function in the engine

### DIFF
--- a/src/utils/mathutils.lua
+++ b/src/utils/mathutils.lua
@@ -363,10 +363,10 @@ end
 ---@return number result # The interpolated angle
 ---
 function MathUtils.lerpAngle(from, to, factor)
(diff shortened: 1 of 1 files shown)


NyakoFox:
Mainly needs formatting fixes (we put spaces between terms) and maybe better variable naming (not a huge fan of `cs` and `sn` -- probably just say `cos` and `sin` assuming thats what they are) but I like the idea!

NyakoFox:
LGTM, thank you!

NyakoFox:
Feel free to open another PR to add yourself to the credits, or tell me what you'd like to be credited as and I'll commit it myself!

PR #526, opened 2026-02-01
Remove self from more Map variables in Event:onRemove

Dobby233Liu:


NyakoFox:
Looks good to me, thanks!

PR #527, opened 2026-02-02
Explain overriding offset_x/y in Text:getCharPosition

Dobby233Liu:
This makes it a bit clearer why getCharPosition is like that imo


NyakoFox:
Those have to exist due to anything that isn't 30 FPS so shake and similar values are not re-calculated every frame.

Dobby233Liu:
I'll just add a comment then

PR #528, opened 2026-02-03
Improve modRequire with caching and library support

Hyperboid:
Sorry if I didn't do a good job explaining this, it's _5AM Again_

Basically the caching is like `package.loaded`, and `modRequire("libraries.mylib.thing")` is the same as `libRequire("mylib", "thing")` but actually with some degree language server support (of course LS will get confused if lib folder name differs from ID)

Motivation: I'm tired of creating and maintaining `globals.lua` files to trick the language server

### DIFF
--- a/src/kristal.lua
+++ b/src/kristal.lua
@@ -1375,6 +1375,8 @@ function Kristal.loadMod(id, save_id, save_name, after)
 
         -- Add the current library to the libs table (again, with the real final value)
         Mod.libs[lib_id] = lib
+        -- Cache the library to be accessible through modRequire/libRequire
+        Kristal.LoadedModScripts["libraries." .. lib_id .. ".lib"] = lib
     end
 
     Kristal.loadModAssets(mod.id, "all", "", after or function()


Hyperfae:
I should explain caching better, shouldn't I? Well.

When `require`-ing a module in Lua, the return value gets cached in `package.loaded`. As such, `require("src.myclass") == require("src.myclass")`.

The same is *not* true with modRequire. modRequire-ing a class file results in an entirely seperate class, which is problematic for functions that take a class as an argument (namely Class:includes and Stage:getObjects). That is, `modRequire("src.myobject")` and `modRequire("src.myobject")` are two _different_ classes that both inherit from Object.

NyakoFox:
I forgot to comment this, but I'm not that sure about a modRequire having the ability to require a library? I'm confused why that's there.

Hyperfae:
Again, it's for language server support. The path paramater in libRequire is the second argument (`libRequire("mylibrary", "src.objects.bannana")`). Meaning, if you configure it to treat libRequire as require, the language server will interpret it as loading a file called mylibrary.lua.

That said, in retrospect, it would've been better to make this PR exclusively focused on the caching. If you want, I can reduce this down to just the first and third commit, removing the library support.

NyakoFox:
I see... I guess I'm just confused by like... does LuaLS not have an import map? Is that only something in EmmyLua?

NyakoFox:
I think I finally understand; didn't know `libRequire` worked like that, but I wonder if we should deprecate it because of that... I'll ask Sylvi and get back to you.

NyakoFox:
Looks good to us, we should deprecate libRequire some time because of this.

PR #529, opened 2026-02-09
Convert Squeak into a registered event

Hyperboid:
Better to use the new system for the template


NyakoFox:
I swear I merged this a week ago

PR #530, opened 2026-02-10
Power Acts accuracy

NellMonell:
In Chapter 4, when we use the dual power act, the text says "* Your SOUL shined its power on RALSEI and SUSIE!", which makes it clear that it saying "Susie" (not in uppercase) in Chapters 1 and 2 was seemingly an oversight. This commit fixes that, now having power acts use always the uppercase writing for the name of the spell caster (and Susie being the exception now instead of Ralsei). Also, in all power acts "SOUL" was written in uppercase too, which for some reason wasn't the case in Kristal. It has been fixed as well.

### DIFF
--- a/src/engine/game/battle.lua
+++ b/src/engine/game/battle.lua
@@ -1673,12 +1673,12 @@ function Battle:powerAct(spell, battler, user, target)
         end
     end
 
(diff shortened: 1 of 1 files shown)


NyakoFox:
LGTM!

PR #533, opened 2026-02-24
Call onWavesDone before setting up exit call in endWaves

Dobby233Liu:
This is so the user can start a cutscene in encounter:onWavesDone and it'd (presumably) work slightly more correctly


Dobby233Liu:
Ok this doesn't truly fix anything, converting to draft

PR #539, opened 2026-03-04
Fixed more bugs related to the SOUL in the main menu

FireRainV:
What was fixed:
* Fixed an issue where the soul wouldn't use the mod's soul color in some menus.
* Fixed an issue where when selecting a save file to load/delete, the soul would have an unusual offset downwards.


FireRainV:
done

PR #540, opened 2026-03-04
Changed the default animation speed of recruits in the menu

FireRainV:
Just to make it more consistent, the default animation speed of recruits in the recruit menu is now the same as the default animation speed of world bullets.

### DIFF
--- a/src/engine/game/common/data/recruit.lua
+++ b/src/engine/game/common/data/recruit.lua
@@ -62,7 +62,7 @@ function Recruit:init()
 
     self.box_gradient_color = { 1, 1, 1, 1 }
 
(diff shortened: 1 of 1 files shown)


NyakoFox:
I'm not sure about this; 0.25 was a weird choice for us (that'd be... 7.5 / 30?)

FireRainV:
do you have a better one? like, what would you put as default?

NyakoFox:
what's wrong with the current one? it's a default either way

FireRainV:
so it's fine? should I just close it?
I just wasn't sure what to put as default, since 4 / 30 is just the animation speed of Virovirokun

FireRainV:
yea, i think i will just go and close it, it doesn't matter anyways

PR #541, opened 2026-03-04
Fixed a crash when attempting to debug select an interactable with a cutscene/script

FireRainV:
I'm not sure what it was suppose to do, it gives a crash as it attempts to use a function as a string.

Did I do it right?


FireRainV:
im not sure if this is a good fix, im closing this

PR #544, opened 2026-03-05
Added an option to log missing sounds

FireRainV:
It took me a while to notice that some stuff I was working on which suppose to play certain sounds didn't actually play because I forgot I removed the file, but the engine never warned me about it.

### DIFF
--- a/src/engine/assets.lua
+++ b/src/engine/assets.lua
@@ -377,6 +377,7 @@ function Assets.startSound(sound)
     else
         Kristal.Console:warn("Sound not found: \"" .. sound .. "\"")
     end
+    ---@diagnostic disable-next-line: return-type-mismatch
     return nil
 end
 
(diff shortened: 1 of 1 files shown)


NyakoFox:
I don't think this should be an option, it should *always* warn. The category is a good idea, but please ask us about large changes like that first...

FireRainV:
done, it should always warn this time.

NyakoFox:
What is the `not OVERLAY_OPEN` check for?

FireRainV:
> What is the `not OVERLAY_OPEN` check for?

I thought it doesn't show warnings if you're in the console window, but I was wrong.
I removed the check.

NyakoFox:
Oh yeah -- why did you remove the diagnostics?

FireRainV:
I didn't know what is was for, sorry.
I thought it was some sort of handler when there's no sound.
I restored it now.

NyakoFox:
Looks good, thanks!

PR #546, opened 2026-03-06
The Overworld Soul is now always located on the player (and some new World functions)

FireRainV:
If the soul character is a follower, battle areas will now work correctly with the soul still being on them.

Before:
Screenshot (67)

After:
Screenshot (66)

### DIFF
--- a/src/engine/game/world/overworldsoul.lua
+++ b/src/engine/game/world/overworldsoul.lua
@@ -75,8 +75,6 @@ function OverworldSoul:update()
 
     local progress = 0
 
-    local tx, ty = sx, sy
-
     if Game.world.player then
         self.x, self.y = Game.world.player:getRelativePos(Game.world.player.actor:getSoulOffset())
         if Game.world.player.battle_alpha > 0 then


NyakoFox:
The way this was supposed to work is that when you enter a battle area, the soul would go from the soul member to the player-controlled character. Keeping the soul on the soul party member seems uh, difficult for actual gameplay and probably shouldn't be how it's done.

(Quite honestly, the entire "soul player" stuff was really strange and we shouldn't have it, but it's too late now.)

FireRainV:
I actually struggled to get what you said to work.
If the bullets are outside a battle area, the soul location would be still on the soul character, so changing it would make it inconsistent.

Tell me if there's any change you want or if is it okay in the first place.

NyakoFox:
Honestly I'm not sure what we should do about this. If you can, maybe just have the soul be on the player always.

FireRainV:
sure, that should be really easy.
(and I spent so much time getting this to work 😢)

FireRainV:
can I keep the new World functions? the getPlayerAndFollowers() and getSoulPartyCharacter()

NyakoFox:
yea

NyakoFox:
could you do `getPartyCharacters()` rather than `getPlayerAndFollowers()`?

FireRainV:
not a good idea, as followers are not always a party member.
this is why I named it like that.

FireRainV:
done, tell me if it's alright.
Screenshot (68)

NyakoFox:
lgtm

PR #552, opened 2026-03-10
A tiny cosmetic change to Noelle

FireRainV:
After looking a bit though Kristal's code, it seems that you use values from the end of the chosen chapter (like Ralsei's title from Chapter 3 being the last one he gets.)

By the end of Chapter 2, Noelle has 100 boldness points, (60 if you check right before she leaves the party.)

So I updated it to make it more consistent.
Screenshot (71)

### DIFF
--- a/data/party/noelle.lua
+++ b/data/party/noelle.lua
@@ -112,9 +112,9 @@ end
 
 function character:getTitle()
     if self:checkWeapon("thornring") then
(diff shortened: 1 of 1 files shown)


FireRainV:
hold on, I just noticed that she's level 3.
Should probably make her level 2 then.

FireRainV:
I mean, the level of the current chapter.

FireRainV:
fixed the spacing, forgot about it

PR #554, opened 2026-03-13
Added "Remove outside of arena" boolean for bullets

FireRainV:
It removes the bullet when it goes off the arena.
This was a requested feature.

### DIFF
--- a/src/engine/game/battle/bullet.lua
+++ b/src/engine/game/battle/bullet.lua
@@ -62,6 +62,8 @@ function Bullet:init(x, y, texture)
 
     -- Whether to remove this bullet when it goes offscreen (Defaults to `true`)
     self.remove_offscreen = true
+    -- Whether to remove this bullet when it goes outside of the arena (Defaults to `false`)
+    self.remove_outside_of_arena = false
 end
 
 --- Get the graze tension for this bullet.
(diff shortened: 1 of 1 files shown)


FireRainV:
should I account for the sprite's size?
I just used the method that Sam Deluxe did for Froggit in Magical-Glass.

AcousticJamm-dev:
well here's what I'm thinking: Say you have a bullet that is masked to the arena and it uses this. The code you're giving is based on a single point where the bullet is; so, if a bullet is big enough, it'd look jarring for it to disappear when its middle leaves the arena.

FireRainV:
> well here's what I'm thinking: Say you have a bullet that is masked to the arena and it uses this. The code you're giving is based on a single point where the bullet is; so, if a bullet is big enough, it'd look jarring for it to disappear when its middle leaves the arena.

Yea, I know that, but I don't know which one is accurate, since Sam Deluxe didn't include it.

NyakoFox:
I don't think this is a good idea; it's not possible to make this feature work well in every case, and a lot of the time will just cause confusion; people enable the boolean, and suddenly the bullet they spawned outside of the arena does not appear. The current implementation relies on bullets being spawned *inside* the arena and leaving it later; something that is not very common. If it *is* desired, manual checks are extremely easy. And if manual checks are extremely easy, there's not really a point in implementing a default system which barely works in the user's favor.

PR #555, opened 2026-03-14
Mercy fix

NellMonell:
Currently, Kristal stops showing the mercy messages if the enemy's mercy reached 100%, which isn't exactly accurate to Deltarune. This change introduces more accurate approach by introducing a second argument to the addMercy() function, that allows to stop showing mercy messages if enemy's mercy is full, while by default they'll be always shown.

### DIFF
--- a/src/engine/game/battle/enemybattler.lua
+++ b/src/engine/game/battle/enemybattler.lua
@@ -440,14 +440,7 @@ end
 
 --- Adds (or removes) mercy from this enemy
 ---@param amount number The amount of mercy being added (or removed, if set to negative)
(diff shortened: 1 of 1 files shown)


NyakoFox:
It's so upsetting that I exclusively tested mercy with Virovirokun in DR because it is like the only enemy with this behavior. I made the mercy system acurrate to specifically Virovirokun and nothing else. I hate Toby Fox

Anyway, I feel like you could just rip out the code instead of adding an argument, and Virovirokun (and others that want the behavior) just check the mercy themselves and avoid calling `addMercy` under the necessary conditions?

NellMonell:
Alright sure, just gimme a minute

PR #556, opened 2026-03-14
Updated Ultimate Heal (all current chapters)

FireRainV:
Other than making Ultimate Heal accurate to chapter 3 and 4, the "best of Susie's ability" will now use the owner's name instead of just Susie.


FireRainV:
I also coded the TP cost in a much better way than Toby did.
Where he checked for each number of usages separately.

FireRainV:
I didn't include OKHeal since it gets replaced by BetterHeal at some point during chapter 4.
Tell me if you want me to include it anyways.

NyakoFox:
I really need you to discuss additions with us (in #engine-dev) before opening PRs for them... this is one of those things that need a lot more discussion before implementing it.

FireRainV:
okay, I will discuss it there next time.
I guess it can stay open for the time being.

FireRainV:
I've updated to make it way easier to choose between what you want, but I'm putting this PR as a draft for now till I figure out if Ally wants it all inside of ULTIMATE HEAL or separated into their own spell files.

FireRainV:
if anyone knows any idea on how I can mange the spell changing states, because I don't want to use a method which could probably get rejected, tell me please.

NyakoFox:
First -- I did say we'd handle this eventually. Second, when you showed interest in doing this, I *did* say that I would make UltimateHeal a separate spell, since it doesn't use any of the logic from the "upgrading" one.

Third, "I coded the TP cost in a much better way than Toby did" needs to show what he actually did, compared to yours, so we can verify they do the same thing.

The `getBattleDescription` change is a good idea.

The changes to `addSpell` and `replaceSpell` are not a good idea. It's messy and ideally everything should be dynamic based on callback functions, NOT some extra data table... or at least if something like that is added in the future, it's a bit more structured.

FireRainV:
you can make changes that you think fits, i will apply them to this PR

he coded it like this:
image

FireRainV:
while my version was doing this:
80 - math.ceil(math.min(chara:getFlag("ultimateheals_used", 0), 15) * (1 / 3))

FireRainV:
as for the logic, the spell works the same way anyway across all of its versions. i just didn't apply any of the benefits from using UltimateHeal over and over. i'm not sure why it's more recommended to separate them.

it will also make it easier to make upgrades for ultimate heal if the user wants to

NyakoFox:
Superseded by #644

PR #560, opened 2026-03-15
Ignore input when overlay is open in GonerChoice

Dobby233Liu:
(cherry picked from commit 6fa3f43dc4f052d0f32c8b8e5b029b358f389b24)


FireRainV:
You think you can apply it everywhere? Like, if you have a custom soul like the yellow soul, it would shoot when you press confirm while on the menu. Maybe something about keypressed/released/up/down?

PR #562, opened 2026-03-16
Added Battle:getSubState()

FireRainV:
There were SET functions for State and SubState, but only a getter function for State.


FireRainV:
I will close this for now and add a bug report, I accidentally forgot to add it to it's own branch.

PR #564, opened 2026-03-21
Added a frame of attacking input lag for accuracy

FireRainV:


NyakoFox (changes requested):

FireRainV:
Is it fine if I keep it as DT and not 1 / 30?
the reason I used DT instead is so when you slow down the game, it won't be truly noticeable and you'll be able to hit the target easily, same goes for playing on high FPS. It's still 1 frame.

I didn't like that change at first because the delay would be noticeable during those scenarios, but DT makes sure it works way better.

NyakoFox:
If the point of this change is for accuracy, then making an opinionated decision to change behaviour outside of 30-fps scenarios is not what you should be doing.

Either way, this isn't really how we'd want it done (timer usage should be very minimal in the base engine & the usages of timers that exist right now aren't great) so I'm going to close this. This is something I wanted to look into eventually anyway because it doesn't seem to be as simple as "add a 1-frame delay".

PR #568, opened 2026-03-21
You'll no longer get healed by default from savepoints in the lightworld

FireRainV:
Accuracy to Deltarune


NyakoFox:
Fixed manually in 86fc863b34b6fffd038f16347ca1df51dd5af2c1, thanks.

PR #571, opened 2026-03-23
Change Discord RPC fallback text from "a mod" to "a project"

Hyperboid:
I don't know when this would come up but still might as well change it


NyakoFox:
Thanks Hyperboid

PR #573, opened 2026-03-25
Fixed Timer.tween to interpolate from the original starting value each frame instead of applying incremental delta updates

FireRainV:
Previously, tweened properties were updated using current = current + delta * ds, which could accumulate floating-point error over time and cause slight misalignment, especially in pixel-perfect UI elements.

The new implementation stores each property’s initial value and evaluates it as start + delta * s every frame, making tweening deterministic and preventing precision drift.

ChatGPT helped me figure out how to fix it, since it's a bit too complex for me.


NyakoFox:
I can't accept this on the basis that we don't want AI-generated code in the engine. If you want, open a bug report and we'll look at it ourselves some time.

PR #576, opened 2026-03-26
Added chain-skipping for text choice box

FireRainV:
In Deltarune & Undertale, when you skip text in the text choice box, it shows everything instantly.

because in this version, the dialogue only appear one after the other, if you press the skip button for just a moment, it won't show everything at once, only the current typing one. In addition, even if you hold the button, it would take a couple of frames to show everything.

I've tested it, it won't run self:update() more than required.

### DIFF
--- a/src/engine/game/common/textchoicebox.lua
+++ b/src/engine/game/common/textchoicebox.lua
@@ -156,6 +156,9 @@ function TextChoicebox:update()
                     end
                 end
             end
+        elseif self:canSkip() then
+            -- Chain skipping
+            self:update()
         end
     end
 end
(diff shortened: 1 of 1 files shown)


FireRainV:
not sure if this is the cleanest solution, since I'm calling self:update() inside the update function, but it works without issues.

NyakoFox:
It just working isn't good enough; our code needs to be clean and maintainable. Putting a manual update call does the trick I'm sure, but it doesn't make sense looking at the code. In the current state of this PR, I can't accept this. If a bug report isn't already open, make sure to open one, and I'll take a look at it myself some other time.

PR #580, opened 2026-03-28
Fixed a very critical inaccuracy

FireRainV:
Kris's dog in the dark stat menu will not disappear upon re-opening the stat menu, only if you hover towards someone else.

In addition, in Chapter 4, it appears in the second slot.

Furthermore, if you load the next map, it will disappear; because of this, I've used "Game.world.party_data" instead of just "self".

Ralsei's dog is probably affected by this as well, but it cannot be seen as he's the third party member, so it was applied to him as well.

### DIFF
--- a/data/party/kris.lua
+++ b/data/party/kris.lua
@@ -176,7 +176,7 @@ function character:onLevelUp(level)
 end
 
 function character:onPowerSelect(menu)
-    if MathUtils.random() < ((Game.chapter == 1) and 0.02 or 0.2) then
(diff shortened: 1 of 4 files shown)


FireRainV:
https://github.com/user-attachments/assets/222c85a3-4a10-4b39-9a56-313e637e2a48

FireRainV:
actually wait, I found a bug where if the party character doesn't exist, it would crash the game

FireRainV:
found a fix

NyakoFox:
This is a very overcomplicated way to do this; in the current state, this PR can't really be merged, sorry.

PR #582, opened 2026-03-29
added option for "blocky" fade in crossFadeTo

HUECYCLES:
This adds an ``options`` table to ``Sprite:crossFadeTo()``, moving ``fade_out`` into it as an option and introducing a ``blocky`` boolean (defaults to ``false``) that enables a "blocky" fade.

### DIFF
--- a/src/engine/objects/sprite.lua
+++ b/src/engine/objects/sprite.lua
@@ -427,6 +427,7 @@ end
 --- Stops the cross-fade on the current sprite
 function Sprite:resetCrossFade()
     self.crossfade_alpha = 0
+    self.crossfade_blocky = false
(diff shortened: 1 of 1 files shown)


NyakoFox:
This looks good, although I'm a little worried about the breaking change to replace an argument with an options table...

HUECYCLES:
i was worried too, i thought about making another function entirely for the blocky fade but it would be. esssentially the same thing and would only change the alpha... and having an additional ``blocky`` argument would also be a breaking change in case people would be using the ``after`` function... so uh

NyakoFox:
Gonna do some followup stuff, ty for the PR!

PR #583, opened 2026-03-29
Added custom color property to Silhouette Event

HUECYCLES:
Adds a custom ``TiledUtils.parseColorProperty(properties["color"])`` to Silhouette.
Defaults to ``{ 0, 0, 0 }``. Properly registered too, I think.
(I tried to add an ``actorcolor`` boolean too for using the actor colors but that didn't end up working so lol)


Hyperfae:
Why is the `color` field annotated as optional? Wouldn't this cause a warning on this line?
```lua
    Draw.setColor(self.color, 0.5)
--                ^ self.color may be nil, according to diagnostic tools
```

HUECYCLES:
hopefully fixed

NyakoFox:
LGTM, thanks!

PR #589, opened 2026-04-09
New face sprites for party members

NellMonell:
Chapters 3 and 4 introduced new face portraits to Susie, Ralsei, and even Noelle. This change will bring those portraits in Kristal


NyakoFox:
This looks good; I'm gonna do a few further changes but thank you for the work!

PR #592, opened 2026-04-15
Error on adding an already added object as a child

Hyperboid:


NyakoFox:
This is one of those things that I hope isn't a breaking change; because if people run into new errors from this, their code has been broken for quite some time...

PR #593, opened 2026-04-17
removed unnecessary usage of PALETTE at Kristal.getSoulColor()

FireRainV:
Just like Game:getSoulColor(), it will directly use red.


NyakoFox:
Yeah this makes sense I think

PR #594, opened 2026-04-17
usage of Kristal.hasAnySaves(mod.id) at MainMenuModList

FireRainV:
it's shorter and better


NyakoFox:
Not sure why it didn't use that originally, was probably just old

PR #597, opened 2026-04-23
Replaced most "mod" uses in comments with "project"

rfrx45:
Replaced all "mod" uses in api-related comments I could find with "project".


NyakoFox:
LGTM, thanks!

PR #598, opened 2026-04-26
Allow BattleCutscene:text() to use overworld Characters as Actors

madrejean:


NyakoFox:
Not sure about this -- ideally, objects from different "states" wouldn't mix. Can I ask why this is desired?

madrejean:
i just thought it would be more convenient this way, after all it's only using it to get the actor. so instead of keeping references in the code to the npc and actor separately, i could just use the npc as a speaker.

NyakoFox:
Denied due to reasons discussed in the Discord server

PR #600, opened 2026-04-28
The new Susie portraits are now default

FireRainV:
If you need the old ones, you can replace them in your mod.


NyakoFox:
Both portraits exist in DR.

PR #602, opened 2026-05-06
Disable noclip on project exit

Hyperboid:
I noticed this was an issue after accidentally leaving noclip enabled and going into sagtuary


NyakoFox:
Do you think the feline was born funny or had to learn

PR #603, opened 2026-05-06
PartyBattler:calculateDamageSimple(amount) will now deal at least 1 damage

FireRainV:
before, it could deal 0 damage.


NyakoFox:
I'm not sure what the point of this change is; this is an unused function in Kristal, and in DR, that condition isn't there. The clamping happens later in the damage calculation, AFTER the "old calculation" function is called. I don't think changing this (unused) function in the base engine does anything.

PR #606, opened 2026-05-10
Added World:checkCollisions()

FireRainV:
Compared to World:checkCollision(), it returns all the objects that the character is currently colliding with at the same time

### DIFF
--- a/src/engine/game/world.lua
+++ b/src/engine/game/world.lua
@@ -469,6 +469,23 @@ function World:checkCollision(collider, enemy_check)
     return false
 end
 
+--- Returns all the inputs `collider` is currently colliding with in the world
+---@param collider      Collider    The collider to check collisions for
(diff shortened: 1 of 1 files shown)


src/engine/game/world.lua
```diff
@@ -475,6 +469,23 @@ function World:checkCollision(collider, enemy_check)
     return false
 end
 
+--- Returns all the inputs `collider` is currently colliding with in the world
+---@param collider      Collider    The collider to check collisions for
+---@param enemy_check?  boolean     Whether to include the enemy collision map in the check
+---@return boolean  collided    Whether a collision was found
+---@return (Object)[]  with        The objects that were collided with
```
NyakoFox:
```suggestion
---@return Object[] collisions The objects that were collided with
```

FireRainV:
applied the changes you did

PR #608, opened 2026-05-17
Many improvements and a fix to the overworld soul

FireRainV:
First, fixed a bug where the soul wouldn't update when the soul character gets changed.

Secondly, the soul will now come from the soul character to the player, but just visually (using love.graphics.translate), meaning that its location and hitbox stays at the position of the player.


NyakoFox:
I'm not sure how long we want to keep the soul priority stuff in the engine... but this is probably fine to have. Thanks for the PR!

PR #609, opened 2026-05-19
math.random() will now be more unpredictable

FireRainV:
before, the sequence of math.random() would always be predictable each time you use it compared to love.math.random().

this should make it way less predictable.


NyakoFox:
neither should be used, use our own random functions

PR #610, opened 2026-05-19
Sprites update

NellMonell:
In Chapter 4, some light world walking sprites were updated. One frame of walking for Noelle, and Susie's walking sprites no longer have two empty pixels at the top (offsets were adjusted due to that). Also, Susie's new unhappy light world walking sprites were added as well


NyakoFox:
This is a bit of a difficult change; Engine accuracy is per-chapter, and by updating these unconditionally, it makes previous chapters less accurate if they did not also get these updates.

Is this something we could lock behind `susieStyle` and potentially `noelleStyle`?

PR #612, opened 2026-05-21
Extract TileLayer preparation into a seperate method from draw

Hyperfae:
Not _100%_ sure regenerateTiles is a good name for this method, but I couldn't think of anything more fitting (and that doesn't imply specifically using a canvas).


NyakoFox:
LGTM -- although I'd just name it `drawTIles`, maybe with a note that it's automatically called when marked dirty.

Hyperfae:
Thing with `drawTiles` is it specifically implies using a canvas, and I intend on rewriting this to use sprite batches which would make the name inaccurate.

NyakoFox:
Sounds good, then!

PR #614, opened 2026-05-23
Fix flag value fallback in Shop:processReplacements

Dobby233Liu:


NyakoFox:
yeah this makes sense... maybe some day we'll stop abusing logic operators so bugs like these stop happening, i wonder how many other things are like this...

PR #615, opened 2026-05-23
Rewrite MirrorArea:drawCharacter

Dobby233Liu:
Rewrites mirrored texture finding logic to handle potential edge cases
better; now:
- Paths in `Actor.mirror_sprites` are explicitly treated relative to the
  actor's current sprite path
- The old frame number is mapped to fit the frame count of the new
  frame set

Makes the drawing of `chara` respect `draw_children_below`/`above`


NyakoFox:
LGTM, thanks!

PR #616, opened 2026-05-24
Fixed crashes related to reloading in certain debug menus

FireRainV:
Simply closing the menu when reloading fixed the crashes.


NyakoFox:
Fixed in 0d0baece3194e13eb08966c300b85812819f33b9. I still don't really like closing the menu, it just needs to be refreshed before a draw is attempted.

PR #622, opened 2026-05-26
Added HeartShine(x, y, properties) effect

FireRainV:
Allows customization the color of the heart at the heart shine.

By default, if the heart is purely white, it will add a black outline to the heart.


FireRainV:
made it easier, cleaner and way more customizable by using properties table instead of parameters.
in addition, the width and height of the object are now based on the background size.

PR #624, opened 2026-05-27
Fixed Playstation Button Layout not displaying

FireRainV:
Kristal didn't recognized my Playstation 4 controller (Dualshock 4) as a Playstation controller.

Love2D called it "PS4 Controller", and for some reason, the check for it was "%f[%a]ps" instead of just "ps".

Screenshot (388)


Hyperfae:
I believe the real fix would be to update the `con` function to use `string.find` directly so it supports patterns again.

NyakoFox:
fixed properly in b2e707d4311e8c23b0bd561e07a1976306823fbe, feel free to verify

FireRainV:
can verify that it works

PR #625, opened 2026-05-28
Fixed inaccuracy with Cell Phone dialogue for Chapter 1.

SweetSylveon:
In Chapter 1 the dialogue for using the Cell Phone in Dark Worlds is "But it doesn't seem to be working." instead of "It's nothing but garbage noise."

### DIFF
--- a/data/items/key/cell_phone.lua
+++ b/data/items/key/cell_phone.lua
@@ -61,7 +61,11 @@ function item:onWorldUse()
         if was_playing then
             Game.world.music:resume()
         end
-        cutscene:text("* It's nothing but garbage noise.")
+        if Game.chapter == 1 then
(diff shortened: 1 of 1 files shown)


data/items/key/cell_phone.lua
```diff
@@ -61,8 +61,12 @@ function item:onWorldUse()
         if was_playing then
             Game.world.music:resume()
         end
-        cutscene:text("* It's nothing but garbage noise.")
+        if Game.chapter == 1 then
+            cutscene:text("* But it doesn't seem to be working."
```
NyakoFox:
```suggestion
            cutscene:text("* But it doesn't seem to be working.")
```

NyakoFox (changes requested):
LGTM otherwise although at some point we should remove all of the direct chapter checks

PR #627, opened 2026-06-03
Added more information for old_item in Shop:drawBonuses()

FireRainV:
Now it's a table that includes:
item: Item
type: ("weapon", "armor")
index: (1, 2)


NyakoFox:
I can't really get behind this as a change; you would only really need this if you're trying to make the menu do other things than what was intended by the code & at that point you should probably just override `drawPartyBonusInfo` (and yes, I do know what your plan was, but I'd recommend rethinking it rather than doing a hacky workaround that also for some reason requires a PR.)

PR #629, opened 2026-06-12
Cleared all spaces from empty lines

FireRainV:
Lines which have no text other than spaces, will now have their spaces removed.


NyakoFox:
🔥🔥🔥🔥🔥🔥

PR #633, opened 2026-06-18
Replaced self.stage with Game.stage and updated onSquished function

SweetSylveon:
There I did it.


NyakoFox:
As I said on Discord - there's no reason to change the default as it's meant to be overridden. However, this is probably a slightly better default.

PR #634, opened 2026-06-20
Updated the entire text format

FireRainV:
I checked them one by one to make sure I didn't replace something that shouldn't be replaced.

Feel free to double-check.


SylviBlossom:
These are definitely good formatting changes, but a codebase reformat like this is better saved for us maintainers to do directly when there's no planned pull requests or in-development changes waiting to be merged - it will cause problems with other PRs and our local changes if we accept this now.

We're closing this since reformatting is difficult to accept in PR form. At some point we may introduce a default formatter config and write a style guide, and run a formatter through the codebase.

FireRainV:
nevermind.
would you want me to give you the wildcards I used to find all of them?

SylviBlossom:
We don't need the wildcards; as I mentioned, preferably we'll set up a formatter sometime so that all of Kristal's code can follow the standards we prefer without having to manually ensure it. Find and replacing all of this stuff manually is crazy impressive work, sorry that we couldn't really accept it

PR #637, opened 2026-06-26
Fixed Kris's and Ralsei's stats and added Noelle's title for Chapter 5

SweetSylveon:


NyakoFox:
I'm confused by the last commit -- Ralsei does lose stats in Chapter 5.

PR #642, opened 2026-07-04
Fixed temporary damage display

HmmNoPls:
The issues are:
- DamageNumber for temporary mercy can start to stretch before it's supposed to, which removes it and prevents future temporary damage displays from appearing
- DamageNumber darkening for negative mercy is now handled through a variable instead of setting the color directly


NyakoFox:
I didn't realize you weren't already in the credits, huh... anyway, looks good to me!

PR #643, opened 2026-07-06
Chapter 5 heals

boatshapedegg:


NyakoFox:
Hey - thank you for the contribution, but we're already working on items and have consumables already finished locally. Unfortunately we will have to close this.

In the future please ask us before putting so much effort into something; it always really sucks when this happens

(Also, we did tell people to refrain from chapter 5 stuff... I guess we only said that on Discord though.)

PR #648, opened 2026-07-08
Move logic for config saving outside of Kristal.setVolume

Goldofleaves:
When editing the master volume, Kristal.setVolume is called every frame, since the logic for config saving is inside Kristal.setVolume, this causes a performance cause the program is writing a to json file every frame.
This pr moves the config saving logic to when the volume is finished setting so it is only called once.


Goldofleaves:
performance issue*

NyakoFox:
oops... LGTM, make sure to add yourself to the credits!

Goldofleaves:
Updated!

Goldofleaves:
i forgot to test the credits code sob now there shouldnt be overlapping

PR #649, opened 2026-07-09
Scrolling console and other console tweaks

jogla-the-wizard:
The console can now be scrolled by pressing Ctrl + Up/Down.
The cursor must be at the top or bottom of the input box, the inputted command will not be deleted by scrolling.
The console may go completely into empty lines, which will all be denoted by a gray `~`.
There is a line counter that will display `: line n of m` where `n` is the current line and `m` the total amount of lines.
Console height is now 16. It was 12 before. This allows to better view the output without having to scroll that much, while still allowing simple multi-line commands, 10 lines from my testing.

I have also added some 'env' commands:
- `help()` - display all `env` functions and basic controls.
- `move(int)` - move the cursor `int` amount of lines. Positive values scroll the console upwards (moves 'cursor' down by one line).
- `moveTo(int)` - moves the cursor to line `int`.
- `resetPos()` - Moves the cursor back to the last line.

Existing commands have not been changed.

Finally I have modified the welcome message to direct the user to the help menu instead of telling them to use `clear()`.


NyakoFox:
This looks great - I was gonna do this eventually but, neverending todo and all that etc etc.

I'd make a few tweaks, though;
- Keep the console 12 tall. 16 is far too large
- Running a script should reset the scroll position
- The line count should probably be somewhere else (and not be an extra line), I'd suggest the right side of the empty space between the lines and the input.

image

Other than those, this looks great!

jogla-the-wizard:
All 3 requests have been fulfilled.
- The console height has been changed back from 16 to 12.
- The scroll position resets when running a script by calling `resetPos()`.
- The line counter has been moved from the left to the right side of the screen. The x position is the same, taking an extra line.
  - To accomplish this, `Console.print` now takes an optional extra parameter, 'align', defaulting to 'left' if absent. If set to 'right', the text will be aligned to the right.

Additionally, I have taken the liberty to replace several instances of the magic number 640 with SCREEN_WIDTH that were scattered around the code.

NyakoFox:
I might do some tweaks to this myself, but this looks good to me! Thank you.

Make sure to add yourself to the credits!

jogla-the-wizard:
Added myself to the credits. Feel free to do whatever you will with my PR.

NyakoFox:
In the future, you should try to rebase your branches rather than merge them; it lets me accept PRs cleanly without having to squash them. :)

PR #653, opened 2026-07-11
Make Susie react if Her ACT defeats the dummy

rfrx45:
if the player manages to get the dummy's health down to one, then ACTs with susie, the dummy gets defeated and nobody reacts to it, so i made Susie say something


SylviBlossom:
This one's kinda true, swearing is DELTARUNE accurate, good enough for me

PR #657, opened 2026-07-14
Fix not being able to heal if invincibility is on

MihBoss96:
Not sure whether it's intentional, but I made a PR just in case.


NyakoFox:
Looks good, make sure to add yourself to the credits

MihBoss96:
There. It's already sorted so you don't have to do it.

NyakoFox:
LGTM, thanks!

PR #663, opened 2026-07-16
Spell's numerical position in addSpell()

NellMonell:
Adds a second argument to PartyMember:addSpell() which allows you to add the new spell in a certain position in the spell list instead of it always being the last one


NyakoFox:
This idea is fine, but it really shouldn't be an optional argument of `addSpell`. Make an `insertSpell` function instead, with the arguments `(index, spell)`.

NellMonell:
Made the change. Is it okay now?

NyakoFox:
Yep, thanks!

PR #665, opened 2026-07-18
Collider restructuring, bounding box checks

SylviBlossom:
Opening a PR for this to have more time for testing / review.

**Changes:**
Colliders have been massively restructured and slightly optimized with bounding box checks
- `collidesWith` has been split into `meetsCollider` / `meetsObject` in most cases, and has been deprecated
- Collision detection between different collider types is now handled with a new CollisionRegistry system
  - Each unique type of Collider (e.g. `"rectangle"`, `"point"`) is now registered to the CollisionRegistry, and Colliders now have a `getColliderType` returning their collider type
  - Collision tests between collider types must now be registered as functions to the CollisionRegistry, optionally symmetrically, and separately for "inner" collisions
- Many more specific changes to Collider functions/fields to list

The Collider changes and new collision registry system is intended to be documented on the wiki soon, for now here's the most important changes for an average codebase:
- `collidesWith` --> `meetsCollider` / `meetsObject`
  - This applies to both Collider and Object, and is a context dependent replacement. `collidesWith` will still work for now although it is deprecated.
- `Battle:checkSolidCollision` --> `Battle:solidMeetsCollider` / `Battle:solidMeetsObject`
- `Collider.parent` --> `Collider:getOwner()` / `Collider:setOwner(object)`
- `Collider.collidable` --> `Collider:isCollidable()` / `Collider:setCollidable(collidable)`
  - `Collider:isCollidable()` also checks if its owner is collidable.
- `Collider.inside` --> `Collider:isInner()` / `Collider:setInner()`
  - `inside` has been renamed to `inner` in every location.
- `PolygonCollider.points` --> `PolygonCollider:getPoints()` / `PolygonCollider:setPoints(points)`
  - You can no longer modify the `points` field directly, due to bounding box optimizations.
- `LineCollider.x`, `LineCollider.y` --> `LineCollider.x1`, `LineCollider.y1`
If you made a custom collider, see the CollisionRegistry class for how to use it, and use the `registerColliderTypes` and `registerCollisions` events respectively.

### DIFF
--- a/main.lua
+++ b/main.lua
@@ -116,6 +116,8 @@ FountainShadowFX = require("src.engine.drawfx.fountainshadowfx")
 GradientFX = require("src.engine.drawfx.gradientfx")
 ScissorFX = require("src.engine.drawfx.scissorfx")
 
+CollisionRegistry = require("src.engine.colliders.collisionregistry")
+KristalCollisions = require("src.engine.colliders.kristalcollisions")
 Collider = require("src.engine.colliders.collider")
 ColliderGroup = require("src.engine.colliders.collidergroup")
 Hitbox = require("src.engine.colliders.hitbox")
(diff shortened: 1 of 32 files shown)


NyakoFox:
Last call for testing; will merge this after 24 hours.

PR #667, opened 2026-07-18
fix weird fullscreen window border bug

Goldofleaves:
Fixes the weird bug with borders as shown in:imageaherrhehaeht5

This happens when you call love.window.setFullScreen(false) when you bootup on fullscreen mode, which is because love.windows.setFullScreen(false) falls back to the original window size before it was initially fullscreened. In the case of bootup, it is exactly the values used in love.conf since the initial Kristal.resetWindow call in love.load handles both fullscreening and editing the window size (which doesnt do anything in this case, since it is also fullscreening).

This pr removes the Kristal.resetWindow call in love.load and edits it to resize the window first (for love.window.setFullScreen to have a correct fallback value), then fullscreen the window.


NyakoFox:
Honestly, I'm not the biggest fan of copy pasting so much code, and I can't reproduce this issue at all. If possible, please provide steps to reproduce the issue, and I'll look into it myself.

PR #671, opened 2026-07-23
Added missing [wait]

SweetSylveon:


SweetSylveon:
Should the interaction with Ralsei in Chapter 1 after tossing it be added?

NyakoFox:
good catch
no, i dont think that should be added, that's very deltarune-specific

PR #672, opened 2026-07-24
Run Oxipng on all PNG Files

Jaduplex:
[Oxipng](https://github.com/oxipng/oxipng) is a useful tool that minimizes the space PNG files take while making absolutely no visual or practical difference. While on individual files it tends to make little difference size-wise, as a unit this can reduce filesize heavily.

For any future use, the batch script I use to run this on all files in a directory and its subdirectories is:
`oxipng -o 6 --strip safe --alpha {} **/*.png`

I hope this provides any benefit.


A1Asriel:
It's a thoughtful idea, but it doesn't help to compress the Git repository because existing versions of assets are permanently saved in history. This PR would only increase the overall size (73.2 MiB -> 76.6 MiB on local test) since more objects would have to be recorded (unrelated, but it also caused a memory leak and nearly made my system crash when I opened the "Files changed" tab lol)
I think a much more rational solution would be to provide your script for creators to run before releasing their game, but saving ~1.3 MiB for the distributable package isn't going to make a big enough difference to justify cluttering the repository, in my opinion.
But, let's see what the maintainers think of this

NyakoFox:
I agree; it'd make cloning take longer, but I do wonder if it'd make Kristal load faster. Kristal's slow loading comes from assets being small individual files; I wonder if making the files smaller would help.

The biggest issue is engine forks and potentially causing merge conflicts with binary files.

NyakoFox:
Unfortunately, this PR doesn't seem to have much benefit, so I'm going to have to close this.

PR #674, opened 2026-07-27
Add FLAC support

A1Asriel:
I just noticed that Kristal doesn't index .flac music and sound files, yet [LÖVE supports this format from version 11.3](https://love2d.org/wiki/Audio_Formats). Since Kristal targets 11.5 and I couldn't even make it run on LÖVE 11.3 or lower anyway, I think it's safe to add FLAC indexing as is, but I could implement additional version check if needed.

### DIFF
--- a/src/engine/loadthread.lua
+++ b/src/engine/loadthread.lua
@@ -123,7 +123,7 @@ local loaders = {
                 pcall(function() mod.logo_data = love.image.newImageData(full_path .. "/logo.png") end)
             end
 
-            local music_extensions = { "mp3", "ogg", "wav" }
+            local music_extensions = { "mp3", "ogg", "wav", "flac" }
             for _, ext in ipairs(music_extensions) do
                 if love.filesystem.getInfo(full_path .. "/preview." .. ext) then
                     mod.preview_music_path = full_path .. "/preview." .. ext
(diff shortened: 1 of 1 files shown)


NyakoFox:
This sounds like a good idea on paper (the more formats, the better) but in practice this isn't really something we want to encourage. FLACs are huge and this may encourage people to include huge lossless FLACs when they shouldn't really do that as music assets.

We already have this issue with mp3 (very lossy and not fit as a game asset) which people keep trying to use; and I've seen people include huge wavs as music, too, instead of using Ogg Vorbis.

(Not to mention the huge amount of tracker formats we support, for some reason...)

A1Asriel:
While true that FLACs are much larger than Vorbis, wouldn't it be better to leave the decision for which format fits their project better to the creator, especially if the upstream engine already allows it? It's worth noting that music is essential to the Deltarune community, some more audiophilic members would prefer to hear the best possible version of the tracks in fangames despite their size. Most devs familiar with this format are most likely aware of the implied drawbacks, and no one is stopping them from distributing a "lite" version with Vorbis-encoded sound assets alongside the original. If I were to release a project with my own music, I would've done exactly that.
So I'm not sure enforcing it so strictly is necessary; besides, there are numerous WAV sound assets that are inferior to FLACs in every way on any device that is capable of running the engine. Allowing WAVs but not FLACs seems even less aligned with your goal.

MP3s, on the other hand, are a real issue due to how popular they are, especially among less IT-literate people. Using them in a fangame is also a choice, albeit a suboptimal one. Though, there might be very specific edge cases that justify it, like using a music file that only exists as an MP3 and not bothering to re-encode it (and losing more quality as a result)

---

I've also been thinking about modifying the `build.py` script with the following improvements: automatic LÖVE download (for standalone releases), AppImage creation support, maybe a shortcut in the debug menu that runs the script automatically, and optional "aggressive" project compression that removes unused assets from the file tree, re-encodes all WAVs (and FLACs, if this PR does get merged) into Ogg Vorbis files, and runs Oxipng for all PNGs (idea from #672) before creating the final package.
Streamlining the process this way would mostly eliminate any concerns about huge projects, as people would be more inclined to do it the "easy" and "right" way.

PR #675, opened 2026-07-27
EnemyBattler: tired functions, more getters

NellMonell:
* The biggest change: added function EnemyBattler:isTired() that checks if the enemy is tired. All checks for enemy.tired have been replaced with that function.
* Added functions onTired() and onAwake()
* Added function hasLowHealth()
* Added getter functions for each text in the getEncounterText() (spareable, tired, low health)

### DIFF
--- a/data/spells/pacify.lua
+++ b/data/spells/pacify.lua
@@ -33,7 +33,7 @@ end
 
 function spell:getCastMessage(user, target)
     local message = super.getCastMessage(self, user, target)
-    if target.tired then
+    if target:isTired() then
         return message
     elseif target.mercy < 100 then
         return message.."\n[wait:0.25s]* But the enemy wasn't [color:blue]TIRED[color:reset]..."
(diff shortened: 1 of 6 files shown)


NellMonell:
Now also added getRareText() and rollRareTextChance(). self.rare_text is also now a thing. This was added due to practically every generic enemy having a rare text the shows up with almost always the same chance. Usually the text is something smell related

SylviBlossom:
The original various encounter text fields are a bit of a design decision we regret, so this PR would be better without the rare text stuff in my opinion. We'd like to encourage overriding `getEncounterText` when making an enemy for complete control over it, rather than supporting a few pre-made conditions for text as fields. Because of this, we'd recommend dropping or reverting your last commit. (That being said, getters for the text fields that exist is a good idea...)

Other than that, it looks good!

NellMonell:
Reverted the rare text. Thanks for letting me know

NellMonell:
If there's any else changes you'd like me to make, just let me know

PR #678, opened 2026-07-31
Remove an extra dot from self.check

MihBoss96:
Removes an extra dot from Box of Heart Candy's self.check.


NyakoFox:
fair enough

PR #681, opened 2026-08-01
Add EnemyBattler:shouldShakeOnHurt(dmg) 

sekalisukarumah-boop:
This adds `EnemyBattler:shouldShakeOnHurt(dmg)` determining whether the enemy should shake when hurt. 
Lmk if it needs adjusting.


MihBoss96:
Maybe I'm wrong, but the function doesn't let you shake it for a certain threshold? I mean, the shake function still uses default values.

sekalisukarumah-boop:
When I said threshold, I meant HP threshold. Should've clarified that, sorry. I could add that, though.

PR #682, opened 2026-08-01
Add 'wish' to the Recruit's Menu

Carsten23455:
Add's the wish field from DR to the recruit menu the same way LIKE and DISLIKE are added (self.wish = "example")


NyakoFox:
This only appears in chapter 5, therefore it'd need a chapter config.

Carsten23455:
alright i'll add it give me a minute

PR #685, opened 2026-08-04
Item sellability and price fixes

NellMonell:
In Deltarune, the item is sellable only if it's price is above 1. Several items had their prices corrected, along with their sellability, following that rule


NyakoFox:
LGTM, thanks!

PR #688, opened 2026-08-11
Makes it impossible to give a "nan" amount of money in the debug options

raisinbrainguy:
Giving a nan amount of money can cause a lot of issues especially since we're adding and subtracting from it in shops _(I think one time it crashed)_ so hopefully this makes it impossible for that to happen!

I don't really mind if you decline this one because 0.1% of the Kristalites will ever do this, I **promise** I'll make actually good pull requests later :P

This is like those insane Instagram conspiracy reels if they were a commit

### DIFF
--- a/src/engine/game/debugsystem.lua
+++ b/src/engine/game/debugsystem.lua
@@ -1229,7 +1229,7 @@ function DebugSystem:registerDefaults()
                 function(text)
                     local money = tonumber(text)
                     if money then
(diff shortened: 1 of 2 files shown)


NyakoFox:
Uhh sure yeah. Add youself to the main menu credits, please

raisinbrainguy:
done 👍 
I also fixed an issue with it that I did not notice even when I was testing it

